Issue #1298, opened 2026-05-31
GeoServices metadata endpoints reject POST (405) — breaks ArcGIS API for Python / ArcGIS Pro metadata hydration


mikemcdougall:
## Problem Summary
Honua's GeoServices REST **metadata** endpoints are GET-only and return **HTTP 405** on POST:

| Endpoint | GET | POST |
|---|---|---|
| `/rest/services/{svc}/FeatureServer` | 200 | **405** |
| `/rest/services/{svc}/FeatureServer/{layer}` | 200 | **405** |
| `/rest/services/{svc}/MapServer` | 200 | **405** |
| `/rest/services/{svc}/MapServer/{layer}` | 200 | **405** |
| `/rest/services/Utilities/Geometry/GeometryServer` | 200 | **405** |

The **ArcGIS API for Python** (`arcgis`) and ArcGIS Pro hydrate layer/service metadata by **POSTing** to these resources. arcgis's `_GISResource._refresh()` does `self._con.post(self.url, {"f":"json"})` first and only falls back to GET on a literal `"Method Not Allowed"` exception — which Honua's bare 405 does not produce in the client. Result: `FeatureLayer.properties` comes back as an **empty PropertyMap**, so the layer never hydrates and **the entire Esri Python SDK (and ArcGIS Pro metadata loading) is unusable against Honua** — even though every operation works fine over GET and `POST .../query` already returns 200.

Real Esri ArcGIS Server accepts **both GET and POST** on all REST resources (POST is how clients send tokens/large payloads); GET-only metadata is non-conformant.

## Why This Matters Now
Surfaced by the `honua-io/honua-esri-compat` harness driving the real `arcgis` 2.4.3 client. This is the single root cause behind "the arcgis SDK returns an empty PropertyMap for Honua services" — a baseline blocker for all Esri-Python-SDK and ArcGIS Pro workflows, gating every downstream operation (query, edit, render).

## Steps to Reproduce
```
curl -s -o /dev/null -w '%{http_code}\n' -X POST \
  http://host/rest/services/browser_compat/FeatureServer/2000 -d f=json    # -> 405
```
Or, with the SDK:
```python
from arcgis.gis import GIS; from arcgis.features import FeatureLayer
fl = FeatureLayer("http://host/rest/services/browser_compat/FeatureServer/2000", GIS())
dict(fl.properties)   # -> {} (empty); con.get(url, {"f":"json"}) returns the full metadata
```
[...]

PR #1329, opened 2026-06-01
fix(geocoding): enable GeocodeServer suggest and batch for Esri SDK certification (#1298)

mikemcdougall:
Related to #1298

## Summary

The Esri SDK certification matrix flagged two GeocodeServer gaps on the config-driven `World/GeocodeServer` (provider: nominatim): `geocode-metadata-POST-to-GET` (the ArcGIS API for Python `batch_geocode` / `geocodeAddresses` failing) and `geocode-provider-suggest` (`/suggest` returning 400). This PR makes both operations work end-to-end with the default nominatim provider, and keeps the genuine "provider can't do it" path intact.

## Changes Made

- **Suggest (`geocode-provider-suggest`)**: nominatim already implements suggest-from-search but reported `SupportsSuggest=false` because `EnableSuggestFromSearch` defaulted off. Defaulted `NominatimProviderConfiguration.EnableSuggestFromSearch` to `true` so `/suggest` works out of the box. Metadata (`currentVersion`/`locatorProperties.SupportsSuggest` and the `capabilities` string) already advertises `Suggest` whenever the active provider supports it, so it now reflects nominatim.
- **Batch parameter name**: Esri clients (ArcGIS API for Python `batch_geocode`, ArcGIS Pro) send the batch payload under the `addresses` parameter (`addresses={"records":[{"attributes":{...}}]}`), but the handler only read `records`, so every batch request 400'd with "records parameter is required". `GeocodingHandler.HandleBatchGeocodeAsync` now reads `addresses` first and falls back to `records`.
- **Batch capability**: nominatim reported `SupportsBatch=false`, so the coordinator returned "no provider supports this operation". Added a default `BaseGeocodeProvider.BatchGeocodeCoreAsync` that fans the batch out to `ForwardGeocodeAsync` (best candidate per input address, in request order); nominatim now advertises `SupportsBatch=true` with a configurable `MaxBatchSize` (default 100). Providers with a native batch endpoint can still override.
- **Tests**: Core unit tests for nominatim capability defaults (suggest+batch) and batch fan-out via forward-geocode; server integration tests for the Esri `addresses` batch parameter returning a `locations` array and metadata advertising `SupportsSuggest`/`SupportsBatch`.

## Testing

- [x] Core unit tests (`Honua.Core.Tests` geocoding) pass, including new nominatim capability/fan-out tests (11/11).
- [x] Release build with `/p:TreatWarningsAsErrors=true` passes (0 warnings, 0 errors).
- [x] Server geocoding integration tests (`GeocodingEndpointTests`) — new `addresses`-parameter batch and `SupportsSuggest` metadata cases added alongside existing suggest/batch coverage.
- [x] Reproduced the before state against the live all-fixes stack (`https://localhost:8443`): metadata POST already 200==GET; `/suggest` returned 400 "no provider supports this operation"; `geocodeAddresses` returned 400 "records parameter is required".

## Gate Impact

## Breaking Changes

None. Behavior is additive: suggest/batch now succeed with the default nominatim provider; explicitly requesting a provider that genuinely lacks a capability still returns the same 400.

---

- [x] Ran scripts/ci/pre-pr-check.sh

> **Validation note:** This change was developed TDD against the GeocodeServer handler and the Core nominatim/`BaseGeocodeProvider` paths. The live `:8443` stack is a separately built deployment, so the before-state was reproduced there directly (suggest 400, batch 400, metadata POST already 200) and the after-state is proven by the Release build (warnings-as-errors, green) plus the Core unit tests and server integration tests exercising the same handler and provider code. The shared multi-agent build lock was heavily contended during this run; targeted geocoding tests and the affected-scope pre-PR check were executed under that lock.

### DIFF
--- a/src/Honua.Geocoding/Features/Geocoding/Abstractions/BaseGeocodeProvider.cs
+++ b/src/Honua.Geocoding/Features/Geocoding/Abstractions/BaseGeocodeProvider.cs
@@ -99,16 +99,49 @@ public abstract class BaseGeocodeProvider : IGeocodeProvider
     }
 
     /// <summary>
-    /// Core implementation of batch geocoding functionality
+    /// Core implementation of batch geocoding functionality.
     /// </summary>
+    /// <remarks>
+    /// The default implementation fans the batch out to <see cref="ForwardGeocodeAsync"/>,
+    /// returning the best candidate for each input address in request order. Providers with a
+    /// native batch endpoint should override this for efficiency. This keeps the Esri
+    /// <c>geocodeAddresses</c> operation working for providers (such as Nominatim) that only
+    /// expose a single-address forward-geocode API.
+    /// </remarks>
     /// <param name="request">Batch geocoding request</param>
     /// <param name="cancellationToken">Cancellation token</param>
-    /// <returns>List of geocoding candidates</returns>
-    protected virtual Task<IReadOnlyList<GeocodeCandidate>> BatchGeocodeCoreAsync(
+    /// <returns>List of geocoding candidates, one per input address in request order</returns>
+    protected virtual async Task<IReadOnlyList<GeocodeCandidate>> BatchGeocodeCoreAsync(
         BatchGeocodeRequest request,
         CancellationToken cancellationToken = default)
     {
-        throw new NotSupportedException($"Provider '{Name}' does not support batch geocoding functionality.");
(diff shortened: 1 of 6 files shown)


mikemcdougall:
Superseded by #1334 — bundled with the other Esri-SDK certification gap fixes into a single PR per maintainer request. Branch preserved.

Issue #418, opened 2026-03-07
Program: Integrated GIS platform roadmap (2026)


mikemcdougall:
## Current State

Honua is already broader than the backlog narrative implies.

The runtime today already includes:
- GeoServices REST: FeatureServer, MapServer, ImageServer, Geometry Service
- OGC API: Features, Tiles, Maps
- OData v4
- WMS 1.3 and WMTS 1.0
- WFS 2.0 read/query surface
- gRPC feature service
- Admin/control-plane APIs for connections, publishing, metadata, deploy coordination, tile operations, observability, and imports
- PMTiles archive generation
- FileGDB, GeoServices, and GeoServer-oriented migration/import workflows

The main gap is no longer raw protocol surface area. The main gap is proof, migration smoothness, client compatibility evidence, and backlog/doc truthfulness.

## Product Thesis

Honua should compete by being the most believable modern GIS server:
- multi-protocol by default
- migration-friendly for GeoServer and adjacent legacy stacks
- cloud-native in data formats and deployment model
- operator-trustworthy in testing, rollout safety, and compatibility evidence

That is a stronger near-term strategy than widening the platform with more speculative feature families.

## Active Wave

These issues should define the active roadmap until they are either shipped or explicitly deferred:

### 1. Backlog and execution hygiene
- [x] honua-io/honua-server#416 Operating cadence: weekly backlog review + scope gate + done/close hygiene
- [x] honua-io/honua-server#485 CI/program audit and gate rationalization

### 2. Public-interface proof
- [x] honua-io/honua-server#469 Epic: public interface quality and client compatibility proof
- [x] honua-io/honua-server#470 shared certification matrix and evidence model
- [x] honua-io/honua-server#461 public-interface coverage gate for WFS and gRPC
- [x] honua-io/honua-server#462 deeper contract coverage in admin, geocoding, and geometry service
- [x] honua-io/honua-server#320 Windows/client certification workflow
- [x] honua-io/honua-server#415 expanded client certification beyond MVP smoke

### 3. Migration moat
- [x] honua-io/honua-server#375 SLD style import/export for GeoServer migration
- [ ] tighten existing GeoServices / GeoServer / FileGDB migration docs and evidence through focused follow-on issues

### 4. Cloud-native vector completion
- [x] honua-io/honua-server#380 umbrella: cloud-native vector formats
- [x] honua-io/honua-server#422 FlatGeobuf import + preview
- [x] honua-io/honua-server#423 GeoParquet import + preview
[...]

PR #1485, opened 2026-06-05
fix(geoservices): GeoServices/WFS bug bundle — edits/query/related-records/WFS/robustness

mikemcdougall:
## Issue Link
Related to #418

## Summary
A single bundle of Esri GeoServices / OGC compatibility bugs found by an SDK-driven bug hunt (real ArcGIS API for Python, @arcgis/core, ArcGIS Maps SDK for .NET) and re-verified against current trunk. Grouped into one PR per the reduce-CI-congestion convention. Each fix has regression tests; the fixes are subsystem-disjoint.

## Changes Made
**FeatureServer edits**
- **Geometry-type enforcement**: a point layer no longer accepts/stores polygon/polyline geometry on add/update — mismatches return a per-feature failure (Esri error code), not silent success.
- **`rollbackOnFailure=true`** now wraps the whole applyEdits (service-level across layers + layer-level) in one transaction and rolls back all edits on any failure (default partial-apply behavior unchanged).
- **`deleteFeatures`** now honors `where` and `geometry`+`spatialRel` selection (not just `objectIds`), reusing the layer-query selection.

**FeatureServer query**
- Accept-and-ignore the standard `returnExceededLimitFeatures` query param (sent by the ArcGIS Maps SDK for .NET) instead of HTTP 400.
- **Clamp `resultRecordCount`** to `maxRecordCount` (with `exceededTransferLimit`) rather than returning 400 when it exceeds the cap.
- **Validate `geometryType`** against the Esri enum and 400 on an unknown value, instead of silently coercing to envelope.

**FeatureServer related records**
- **`queryRelatedRecords`** now resolves related records for any origin object-id magnitude (previously empty for high object-ids).

**WFS 2.0**
- Expose **geometry** for all FeatureServer-backed feature types (a point layer previously advertised/returned no geometry in DescribeFeatureType/GetFeature).
- Emit `xsi:nil` for **null** attribute values instead of an empty string.

**Robustness / security**
- An **oversized request body** now returns a clean 413 (no HTTP 500, no leaked internal form-limit debug text).
- The SQL-injection content filter no longer rejects legitimate `--` / `/* */` sequences in feature **attribute values** (applyEdits payloads are parameter-bound); where/SQL-expression injection protection is unchanged and still enforced (regression-tested).

## Testing
- [x] Architecture tests pass

Each fix has regression tests across `Honua.Protocols.GeoServices.Tests`, `Honua.Protocols.OgcClassic.Tests`, and `Honua.Server.Tests`. Full-solution Release `-p:TreatWarningsAsErrors=true` build **0/0**. Affected unit + integration classes pass locally (the related-records high-oid integration coverage is self-contained — it inserts its fixture via applyEdits rather than touching the shared seed, so there's no shared-fixture blast radius). Bugs were reproduced live against a running server, then locked in with the regression tests above.

## Gate Impact
- [ ] None — no gate impact

## Docs or Contract Impact
- [ ] OpenAPI spec changed
- [ ] Documentation updated

Behavior changes align Honua with documented Esri/OGC behavior (param acceptance/clamping, edit validation, WFS geometry, error shaping). No incompatible response-shape changes.

## Release/Deploy Impact

## Breaking Changes
None — fixes are additive/conformance corrections (a previously-rejected param now accepted, invalid edits now rejected per spec, WFS now exposes geometry, oversized body returns 413 not 500).

---

## Pre-PR Checklist


mikemcdougall:
Update: dropped the **resultRecordCount clamp** from this bundle. `LimitsEnforcementTests` showed that `maxRecordCount` is a deliberate operator **governance hard-limit** (returns 400 "exceed configured limits" when exceeded), not the Esri soft clamp — so the adversarial "should clamp" finding was a false positive. Reverted the clamp + its tests; the governance 400 behavior is unchanged. The other query fixes (accept `returnExceededLimitFeatures`, validate `geometryType`) remain. Also repurposed two pre-existing `GeometryValidationTests` that had been asserting the bug (a point layer accepting polygon/linestring edits) to assert the corrected per-feature rejection.

PR #1502, opened 2026-06-05
fix: Postgres file-import staging table + metadata v2 publish bootstrap (#1400, #1395)

### LINKED ISSUE #1395: Metadata v2: layer publish 23505 on idx_metadata_v2_services_name when no snapshot is activated but sidecar rows exist
## Summary

Publishing a layer can fail with a raw Postgres `23505 duplicate key value violates unique constraint "idx_metadata_v2_services_name"` when the Metadata v2 store is in an inconsistent state: no activated current snapshot (`GetCurrentAsync` throws "No Metadata v2 snapshot has been activated for environment '…'.") **and** the `honua.metadata_v2_services_idx` sidecar already has rows.

## Where

- `Honua.Postgres.Features.Metadata.PostgresMetadataV2GraphStore.SaveAsync` → `RefreshSidecarsAsync` inserts the graph's services into `honua.metadata_v2_services_idx` (unique index `idx_metadata_v2_services_name` on `(environment, revision, lower(name))`).
- Reached via `PostgreSqlLayerPublishingService.UpsertPublishedLayerMetadataV2Async` → `LoadCurrentOrEmptyGraphAsync` (honua-server#1341), which on a missing snapshot starts from an empty graph and forces a first write. When the sidecar tables are not actually empty (orphaned rows from prior partial writes / a shared DB), the write collides.

## Impact

- The admin layer-publish (`POST /api/v1/admin/connections/{id}/layers`) returns a redacted 500; the operator sees no actionable reason.
- Same root cause as the long-standing "No Metadata v2 snapshot has been activated" 500 on `GET /api/v1/admin/services`.

## Repro (observed)

A server DB that has `honua.metadata_v2_services_idx` rows but no `honua.metadata_v2_current` row. Any layer publish then 23505s. Testbed workaround: `TRUNCATE honua.metadata_v2_* CASCADE` and restart the server (to clear the in-memory snapshot cache), after which publishes bootstrap cleanly.

## Suggested fix

- Reconcile on bootstrap: when `LoadCurrentOrEmptyGraphAsync` starts from empty because no snapshot is activated, `RefreshSidecarsAsync` should clear stale sidecar rows for the environment (not just the target `(environment, revision)`), or the store should activate/repair a current snapshot from existing sidecar state.
- Add service-/resource-name uniqueness to `MetadataV2GraphValidator` so a graph that would violate the index fails with a clear validation error instead of a raw 23505.

## Related

Surfaced while building provider-aware connection testing + the console publish-layer workflow (branch `feat/provider-aware-connection-test`). The separate non-transactional-publish orphan ("Layer already exists") is fixed on that branch (Metadata v2 projection now runs before the layer-txn commit); this snapshot/sidecar inconsistency is the remaining piece.

### LINKED ISSUE #1400: File import: streamed upload fails with 42P01 'relation public.imported_<table> does not exist' (staging table not visible to load)
## Summary

`POST /api/v1/admin/import/upload` fails for a valid geospatial file: the response is HTTP 200 with `{"success":false,"errorMessage":"Import failed.",...}`, and the server log shows:

```
Npgsql.PostgresException 42P01: relation "public.imported_<table>" does not exist
  at Honua.Postgres.Features.FileImport.StreamingFileImportService...
```

The import creates a staging table named `imported_<table>` (`StreamingFileImportService.Sql.cs` `GetAllowedTableName` → `"imported_" + sanitized`, and `CreateTableAsync` runs `CreateImportTableSql`), but the subsequent data load can't find `public.imported_<table>`.

## Likely cause

A connection/transaction mismatch between staging-table creation and the load: the batch load opens a transaction at `RepeatableRead` (`StreamingFileImportService.Batch.cs`), so if the `CREATE TABLE` runs on a different connection (or commits after the RepeatableRead snapshot is taken), the load's snapshot does not see the staging table. (Same class of non-transactional/connection-boundary issue as the layer-publish orphan fixed on branch `feat/provider-aware-connection-test`.)

## Repro

```
curl -H "X-API-Key: <key>" \
  -F "file=@points.geojson;type=application/geo+json" \
  -F "TableName=demo" -F "TargetSchema=public" \
  http://<server>/api/v1/admin/import/upload
# -> {"success":false,"errorMessage":"Import failed.",...}; log: 42P01 relation "public.imported_demo" does not exist
```

## Suggested fix

Create the staging table and run the COPY/load on the same connection within the same transaction (or create + commit the staging table before opening the RepeatableRead load transaction), and surface the underlying Postgres error in `errorMessage` instead of the generic "Import failed.".

## Related

Surfaced building the console file-import UI (upload + progress + format validation) on honua-console branch `feat/operate-connection-create-test`. Format detection (`/import/formats`), multipart parsing, and format validation all work; only the staged ingest fails.

mikemcdougall:
## Issue Link
Closes #1400
Closes #1395

## Summary
Fixes two Honua.Postgres data-path bugs surfaced while building the console connection/import workflows.

- **#1400** — file-import streamed upload failed with `42P01 relation "public.imported_<table>" does not exist`: the batch load opened its transaction at RepeatableRead before/separately from the staging-table `CREATE`, so the load's snapshot could not see the staging table. The generic `"Import failed."` message also hid the real Postgres error.
- **#1395** — layer publish failed with raw `23505 duplicate key ... idx_metadata_v2_services_name`. Investigation showed the `UNIQUE` index on `lower(name)` was itself the defect: one logical service is intentionally projected across five protocol facets (feature/map/image/ogc/stac) that share a display name with distinct service ids. Adding a uniqueness validator (the originally-suggested fix) would reject valid graphs and broke 11 LayerPublishing tests, so the index is relaxed instead, plus bootstrap reconciliation of orphaned sidecar rows.

## Changes Made
- `StreamingFileImportService.Streaming.cs` — always create the staging table before the load (removed the `OverwriteExisting` gate) so it is visible to the load transaction.
- `StreamingFileImportService.cs` — catch `PostgresException` and surface `MessageText` in `errorMessage` instead of the generic "Import failed."
- `PostgresMetadataV2GraphStore.cs` — bootstrap reconciliation: when no current snapshot is activated, clear stale sidecar rows for the whole environment (all revisions) before the first write.
- `Migrations/031_CreateMetadataV2Snapshot.sql` — services-name index changed from UNIQUE to plain (fresh DBs).
- `Migrations/047_RelaxMetadataV2ServiceNameIndex.sql` (new) — drops + recreates the index non-unique on existing DBs.

## Testing
- [x] Architecture tests pass

New `StreamingFileImportStagingTableTests` (2), `PostgresMetadataV2GraphStoreFreshDbTests` (bootstrap-orphan), `MetadataV2ModelTests` same-named-services-valid. Ran (Docker available): import+metadata Postgres tests (11), MetadataV2ModelTests (21), full LayerPublishingIntegrationTests + exception mapping (28), admin services/metadata-v2 server tests (5), migration-safety architecture tests (11). All green. Release build `TreatWarningsAsErrors=true`: 0 warnings; `dotnet format --verify-no-changes`: clean.

## Gate Impact
- [ ] None — no gate impact

## Docs or Contract Impact

## Release/Deploy Impact
- [x] Requires database migration
- [ ] None — standard merge-and-release flow

Adds migration `047_RelaxMetadataV2ServiceNameIndex.sql` and relaxes the index in `031`; both run on boot.

## Breaking Changes
None

---

## Pre-PR Checklist


mikemcdougall:
Addressed both review items in 122edcac:
- The PostgresException path no longer relays `ex.MessageText`. Client now gets a sanitized, SQLSTATE-coded message (`Import failed: <friendly reason> (SQL state <code>)`) while full detail stays in the server log — no relation/function/schema/constraint names leak.
- The staging-table test now restores the shared `honua` import helper functions in its `finally` block, so the `Database` collection fixture is left intact regardless of execution order. Assertions updated to verify sanitization (no internals, includes SQLSTATE).

PR #1614, opened 2026-06-12
fix: scope edit licensing to FeatureServer

mikemcdougall:
# Pull Request

## Issue Link
Fixes #1591

## Summary
Scopes Pro edit licensing so FeatureServer editing remains gated while equivalent open protocol edit paths are not unintentionally blocked for Community deployments.

## Changes Made
- Updates the licensing catalog and development entitlement behavior for FeatureServer edit gating.
- Moves edit-gate enforcement out of OGC API, OData, WFS, and gRPC edit paths where the issue requires Community edit compatibility.
- Updates capability manifest and protocol tests around the FeatureServer-only edit entitlement boundary.

## Testing
- [x] Unit tests added/updated
- [x] Integration tests added/updated
- [ ] Architecture tests pass
- [ ] Manual testing performed

Validation status:
- Worker validation was still in progress when the branch was inspected.
- A bounded focused `CapabilityManifestEndpointTests` run was active in the worker worktree after an earlier broader focused server test exceeded practical local runtime.
- `scripts/ci/pre-pr-check.sh` has not completed for this branch.

## Gate Impact
- [x] PR gates (build, test, governance)
- [ ] Nightly gates (conformance, performance, security)
- [ ] Release gates (packaging, publishing)
- [ ] Deploy gates (promotion, post-apply validation)
- [ ] None — no gate impact

## Docs or Contract Impact
- [ ] OpenAPI spec changed
- [ ] Protobuf/gRPC contract changed
- [ ] Control plane SDK surface changed
- [x] Documentation updated
- [ ] None — no docs or contract impact

## Release/Deploy Impact
- [ ] Requires coordinated release across repos
- [ ] Requires database migration
- [ ] Requires infrastructure changes
- [ ] Requires environment variable or secret changes
- [x] None — standard merge-and-release flow

## Breaking Changes
None

---

## Pre-PR Checklist
- [ ] Ran `scripts/ci/pre-pr-check.sh` and all checks passed
- [x] Commit messages follow conventional format: `type: description (#issue)`
- [x] PR title matches main commit message
- [x] Issue number linked above
- [x] Tests added for new functionality
- [x] If protocol/auth behavior changed: updated compatibility contract
- [ ] If breaking admin/control-plane API changes: updated migration guide
- [ ] If breaking gRPC/proto wire changes: confirmed with explicit review

Issue #1575, opened 2026-06-09
bug: IRoleStore may resolve to in-memory singleton instead of durable PostgresRoleStore


mikemcdougall:
## Problem Summary

`IRoleStore` has two registrations and it is unclear which wins at runtime:

- `src/Honua.Server/Program.cs:447` registers `InMemoryRoleStore` **unconditionally** as a **Singleton** (comment ~line 442: "in-memory implementations until #496, #498, #355 land").
- `src/Honua.Postgres/ServiceCollectionExtensions.cs:213` registers a durable `PostgresRoleStore` (`src/Honua.Postgres/Features/Authorization/PostgresRoleStore.cs`) as **Scoped**, when the Postgres provider is active.

If the in-memory singleton wins when `DataSource:Provider=postgres`, RBAC role/permission grants are **not persisted** and reset on every process restart. There is also a service-lifetime mismatch (Singleton vs Scoped) for the same interface.

## Why This Matters Now

RBAC is a primary security control. Silent loss of grants on restart (or in a multi-replica deployment where each node holds its own in-memory store) is a correctness and security issue. Surfaced during the Phase 1 security compliance review (honua-io/honua-compliance#17) while grounding the OWASP ASVS V4 / CIS Control 5 self-assessment.

## Steps to Reproduce

1. Configure `DataSource:Provider=postgres` with a real connection.
2. Create a role/permission grant via the admin RBAC endpoints.
3. Restart the server process.
4. Observe whether the grant survives (and whether `PostgresRoleStore` or `InMemoryRoleStore` is resolved for `IRoleStore`).

## Expected Behavior

When a durable provider is active, `IRoleStore` resolves to `PostgresRoleStore`; grants persist across restarts and are shared across replicas. Service lifetimes are consistent.

## Actual Behavior

Unverified — needs confirmation of DI resolution order. `Program.cs:447` registers the in-memory singleton unconditionally; the Postgres scoped registration is added during provider selection (later in `Program.cs`, around the `DataSource:Provider` read). Last-registration-wins vs the Singleton/Scoped mismatch must be checked.

## Acceptance Criteria

- [ ] Determine which `IRoleStore` is resolved at runtime for each provider (postgres / others / none).
- [ ] When a durable provider is active, `PostgresRoleStore` is used; the in-memory store is only the dev/no-DB fallback.
- [ ] Resolve the Singleton-vs-Scoped lifetime mismatch for `IRoleStore`.
[...]

Issue #1567, opened 2026-06-09
chore: repo-wide assembly grade sweep — bring all assemblies to A


mikemcdougall:
### Problem Summary
Track a repo-wide quality sweep that reviews every Honua assembly (34 src + 20 test) against the architecture rubric with a style/maintainability/performance lens and fixes each to an A grade.

### Why This Matters Now
Periodic assembly-level hygiene keeps warnings-as-errors clean, removes duplicated/dead code, fixes sync-over-async, and closes XML-doc and ADR-0011 test-attribute gaps before they accumulate.

### Acceptance Criteria
- [ ] Every assembly reviewed and at grade A (no blocking violations, minimal warnings).
- [ ] Solution builds clean under TreatWarningsAsErrors; dotnet format clean.
- [ ] Core unit tests and Fast-tier integration tests pass.
- [ ] Cross-project follow-ups filed (#1564, #1565, #1566).

### Affected Repo(s)
honua-server

### Gate Tier Impact
PR gates (build, test, governance)

### Release/Deploy Impact
None — standard merge-and-release

### Non-Goals
No behavior changes, no public API/namespace changes, no dependency/package changes.

PR #1662, opened 2026-06-13
fix: recover orphaned quality-sweep follow-ups (#1567/#1599) + reconcile OData with trunk (#1593)

### LINKED ISSUE #1593: tech-debt: coordinated follow-ups from repo-wide quality sweep (cross-cutting + geodesy/standards reviews)
## Problem
The repo-wide quality sweep (PR #1563 / #1567) validated and fixed 253 findings in-project, but 26 confirmed findings require coordinated changes (cross-project interfaces, wire contracts, shared persistence design, or shared SQL-translator semantics) that were out of scope for a single-PR sweep.

## Why now
These are verified-real findings from the cross-cutting + geodesy/standards review waves; several are correctness or security relevant (idempotency claim handling, replica sync last-write-wins, OData null-ordering semantics, entitlement gates on geocoding).

## Findings requiring coordination
- [x] **Cancel endpoint can overwrite a worker's terminal Completed state (TOCTOU, no compare-and-set)** (`src/Honua.Server/Features/Admin/OperationsProgressEndpoints.cs:794`, medium, contract)
  - The correct fix is compare-and-set semantics on IUniversalProgressStore (a Honua.Core cross-project interface, explicitly tracked as a follow-on in the code's own TOCTOU comment at OperationsProgressEndpoints.cs:562-570); the suggested no-CAS palliative (post-write re-read) cannot detect the harmful Cancelled-overwrites-Completed ordering and the handler already does three pre-write re-reads plus worker-owned terminal-state delegation.
- [ ] **Cloud-demo write-token guard only matches GeoServices REST paths; other protocol write surfaces bypass it** (`src/Honua.Server/Features/CloudDemo/CloudDemoEndpointGuards.cs:116`, medium, needs-coordination)
  - Path-matching more protocols in the middleware would be a fragile blacklist; the real fix is a guard in the shared edit/transaction pipeline (canonical IFeatureWriter/edit authorization in Honua.Core/shared server infrastructure, keyed on the seeded cloudDemo guardedWrites flag), which is central shared code outside my file set and spans multiple protocol slices — needs its own coordinated change with endpoint tests per surface.
- [x] **Idempotency replay returns cached 'failed' responses flagged Retryable=true, making same-key retries impossible** (`src/Honua.Server/Features/Forms/FormSubmissionService.cs:365`, medium, needs-coordination)
  - Correct fix (expire/delete the idempotency claim for transient 'failed' outcomes so the same key can re-execute) requires adding a delete/reopen operation to IFormPackageStore in Honua.Core and its PostgresFormPackageStore implementation in Honua.Postgres - both outside this fixer's allowed file scope; a purely local change to ResolveIdempotencyReplay cannot enable re-execution because CreateSubmissionAsync would still conflict on the existing record.
- [ ] **Geocoding Pro/Enterprise entitlements (geocoding.forward/reverse/batch/failover) are never enforced at runtime** (`src/Honua.Server/Features/Geocoding/GeocodingEndpoints.cs:35`, medium, needs-coordination)
  - The finding is real (no LicenseGate usage anywhere in the geocoding path), but adding RequireEntitlement gates would 402 the entire existing GeocodingEndpointTests suite, which runs unlicensed (Community) via the default TestWebApplicationFactory and asserts 200s; the fix must land together with test-suite entitlement grants (TestLicenseEntitlementService) and new 402 gate tests, and tests are outside this fixer's editable scope.
- [ ] **Feature-lock holder identity is client-supplied, not bound to the authenticated principal** (`src/Honua.Server/Features/Collaboration/FeatureLocks/FeatureLockModels.cs:29`, medium, contract)
  - Binding holder identity to the authenticated principal changes client-visible API semantics that integration tests codify: FeatureLockEndpointsTests claims as 'alice' then 'bob' from the SAME authenticated client and asserts a HeldByOther conflict, and asserts the response echoes the client-supplied holderId. Server-derived identity would invert those outcomes; needs a coordinated design change plus test updates (mitigated today by the fail-closed FailClosedFeatureLockAuthorizer, which forbids all mutations).
- [ ] **Fallback mode promotes every instance to leader on Redis failure (split brain)** (`src/Honua.Server/Features/Infrastructure/Coordination/RedisDistributedLeaderElection.cs:139`, medium, needs-coordination)
  - The fail-open fallback is explicitly test-enforced: RedisFailure_FallsBackToLocalMode asserts that a Redis exception with allowFallback=true promotes the instance to leader ('should fallback to local leadership on Redis failure'). Making allowFallback deployment-gated or failing closed in multi-node profiles requires config plumbing plus updating those tests (out of scope); note the class currently has no production construction site in src/ (tests only), so the split-brain risk is latent.
- [ ] **BCL exceptions and name-substring matching map server faults to client error codes** (`src/Honua.Server/Features/Infrastructure/Middleware/GlobalExceptionMiddleware.cs:234`, medium, needs-coordination)
  - Mechanism partially refuted: ExceptionCategorizer's StatusCode/Message are inert (GenerateOptimizedErrorResponse is never called; the wire response always goes through StandardErrorResponse.FromException). The real ArgumentException/ArgumentNullException->400 mapping lives in ExceptionMapper in src/Honua.Hosting/Features/Models/ExceptionMapper.cs, outside src/Honua.Server/ which I am restricted to; changing it is also a wire-behavior change shared by all protocols and needs a coordinated fix there.
- [ ] **Anonymous /v1/spec surface lets unauthenticated callers start background compute** (`src/Honua.Server/Features/Spec/SpecEndpoints.cs:41`, high, needs-coordination)
  - The primary remediation (RequireAdminAuthorization on /v1/spec and gRPC SpecService) is an intentionally documented early-access decision deferred to TODO(#1144); flipping it would break the SDK/test-harness transport per the in-code rationale and requires updating Honua.Architecture.Tests.EndpointAuthorizationGuardTests (tests/ is off-limits to this fixer). The 'at minimum' mitigation WAS applied: MaxConcurrency is now clamped to 16 in both the REST (SpecEndpoints) and gRPC (HonuaSpecService) adapters via shared SpecEndpoints.MaxApplyConcurrency (reported under server-misc-12); bounding total concurrent apply runs lives in SpecApplyOrchestrator (src/Honua.Geometry, outside this project).
- [ ] **Schedule publications persist cron WorkflowDefinitions durably in Redis while the owning publication/package live only in memory** (`src/Honua.Server/Features/WorkflowPackages/WorkflowPackageService.cs:268`, medium, needs-coordination)
  - Real bug, but the correct fix is either a durable (Redis-backed) IWorkflowPackageStore (new persistence design for packages/versions/publications) or a scheduler-side orphan reaper in WorkflowSchedulerBackgroundService that deletes/skips 'workflow-package:'-prefixed definitions with no resolvable publication — the latter couples the Orchestration feature to WorkflowPackages, has multi-node semantics (per-node in-memory publications vs shared Redis definitions) where a naive restart-time purge would kill definitions owned by a live peer, and is too risky to land without integration tests. Added an XML-doc warning on InMemoryWorkflowPackageStore documenting the non-durable store and the orphaned-cron-definition hazard.
- [x] **synchronizeReplica replica-state update is an unguarded read-modify-write** (`src/Honua.Protocols.GeoServices/FeatureServer/FeatureServerRequestHandlers.Replication.cs:893`, medium, needs-coordination)
  - Validated as real (last-write-wins replicaStore.SetAsync after unguarded read in HandleSynchronizeReplica), but a correct fix requires optimistic concurrency (version/etag) on ReplicaRecord in Honua.Core's public IReplicaRepository or a shared distributed-lock primitive that does not exist in shared infrastructure; an in-process per-replicaId lock inside the protocol adapter would silently fail to guard multi-node deployments (the documented scale profile), so this needs a coordinated Core-side change.
- [ ] **geoservices-misc-9** (`?:?`, ?, needs-coordination)
  - Correct fix needs a specific duplicate-version domain exception (none exists) declared in Honua.Core.Abstractions and thrown by PostgresVersionManager - both outside src/Honua.Protocols.GeoServices/; a message-sniffing local workaround would be worse.
- [x] **$orderby null placement inverted: PostgreSQL default NULLS LAST (ASC) violates OData null-before-values rule** (`src/Honua.Postgres/Features/FeatureStore/Services/FeatureQueryBuilder.Ordering.cs:28`, high, needs-coordination)
  - Fixing OData null placement (NULLS FIRST for ASC) requires a per-protocol null-ordering flag on OrderByClause or FeatureQuery, both defined in Honua.Core.Abstractions which is outside the src/Honua.Postgres/ boundary.
- [ ] **GetEffectivePermissionsAsync issues an uncached, non-sargable DB query on every authorization decision** (`src/Honua.Postgres/Features/Authorization/PostgresRoleStore.cs:219`, medium, needs-coordination)
  - PostgresRoleStore is registered as Scoped, so a field-level cache would be discarded each request. A proper short-TTL cache requires either a singleton caching decorator on IRoleStore or a shared IMemoryCache, both requiring changes outside src/Honua.Postgres/ (DI registration in Honua.Postgres/ServiceCollectionExtensions.cs could host a decorator but no package reference exists for IMemoryCache and no-new-packages rule applies).
- [x] **ICrsDetectionService drops cancellation before PostGIS queries reached from the GeoServices request path** (`src/Honua.Postgres/Features/FileImport/CrsDetectionService.cs:121`, medium, needs-coordination)
  - Adding CancellationToken to ICrsDetectionService methods requires modifying the public interface in Honua.Core/Features/Infrastructure/Abstractions/ICrsDetectionService.cs and updating callers in Honua.Protocols.GeoServices (SpatialReferenceResolver) and elsewhere — outside the src/Honua.Postgres/ boundary.
- [x] **$filter `ne` translated to SQL `<>` silently excludes NULL rows (3VL vs OData 2VL)** (`src/Honua.Postgres/Queries/Filters/PostgresSqlFilterTranslator.cs:79`, high, needs-coordination)
  - Real OData v4.01 conformance gap, but PostgresSqlFilterTranslator is a shared sink for CQL2/FES/Esri SQL where-clauses which mandate SQL 3VL for '<>'; the expression tree carries no protocol marker, so the correct fix is in ODataFilterParser.NormalizeComparison (src/Honua.Geometry, outside this project) — e.g. rewrite 'x ne y' as Or(NotEqual, IsNull(x)) — or a dialect flag in the Core expression model, mirrored in the MySql/DuckDB translators.
- [x] **geo.intersects evaluated planar (geometry) on geographic layers while geo.distance is geodesic (geography)** (`src/Honua.Postgres/Queries/Filters/PostgresSqlFilterTranslator.cs:214`, medium, needs-coordination)
  - Real OData Edm.Geography semantics gap, but TranslateSpatial serves CQL2 S_INTERSECTS / WFS Intersects whose planar-in-CRS evaluation is CITE-validated (952/952); blanket geography routing for geographic layers would also break predicates PostGIS geography does not support (Contains/Within/Crosses/Touches/Overlaps). Needs a protocol-level geodesic marker in the shared expression model (Honua.Core/Honua.Geometry) before the Postgres translator can route OData-sourced Intersects through ::geography.
- [x] **Synchronous Redis I/O (ScriptEvaluate/StringSet/KeyExists) on request threads** (`src/Honua.Hosting/Features/Caching/DistributedCacheRefreshCoordinator.cs:450`, medium, needs-coordination)
  - NotifyDistributedInvalidation and ReleaseDistributedClaim are already async (fire-and-forget ScriptEvaluateAsync). Remaining sync calls are in TryClaimDistributedRefresh (StringSet), TryClaimDistributedWriteBack (KeyExists + ScriptEvaluate), and WasDistributedInvalidated (KeyExists). Fixing them correctly requires changing the ICacheRefreshCoordinator interface in Honua.Core (TryEnqueueRefresh, TryClaimWriteBack, WasInvalidated return bool synchronously) - a cross-project API change.
- [x] **Geometry CRS transform on write paths makes one PostGIS round trip per vertex** (`src/Honua.Protocols.OgcApi/Features/Services/OgcFeaturesGeometryServices.cs:590`, medium, needs-coordination)
  - TryTransformCoordinateSequenceAsync still does per-vertex TransformPointAsync awaits. Fixing requires adding a batch transform method to ICoordinateTransformService in Honua.Core (outside editable scope) or rewriting the edit pipeline to use ST_Transform server-side.
- [x] **Batch transaction prepares and maps up to 1000 operations with sequential per-operation DB round trips** (`src/Honua.Protocols.OgcApi/Features/OgcFeaturesTransactionHandler.cs:1299`, medium, needs-coordination)
  - PrepareBatchOperationsAsync and MapBatchEditResultsAsync still issue per-operation DB round trips. Fixing requires new multi-resolve capabilities on the shared query pipeline in Honua.Core (outside editable scope).
- [x] **If-Match precondition check is not atomic with the edit (TOCTOU on optimistic concurrency)** (`src/Honua.Protocols.OgcApi/Features/OgcFeaturesTransactionHandler.cs:403`, medium, needs-coordination)
  - ETag check is not atomic with the write. Enforcing precondition inside the transaction requires changes to the edit pipeline in Honua.Core/Honua.Postgres (outside editable scope).
- [ ] **Stored queries kept in per-process static state break multi-node deployments** (`src/Honua.Protocols.OgcClassic/Wfs20/Services/Wfs20Handler.StoredQueries.cs:46`, medium, needs-coordination)
  - Persisting managed stored queries across replicas requires a shared Redis/metadata store — an architectural change outside src/Honua.Protocols.OgcClassic that touches infra-layer contracts.
- [x] **Change-set PATCH merge and ETag precondition are checked against a read taken before the deferred batch write (TOCTOU / lost update)** (`src/Honua.Protocols.OData/Features/Services/ODataBatchHandler.ChangeSet.cs:233`, medium, needs-coordination)
  - TOCTOU/lost-update fix requires compare-and-set enforcement inside the feature writer's ApplyEditsAsync transaction, which lives in IFeatureWriter/Honua.Core/Honua.Postgres outside this project.
- [ ] **stac-conformance-8** (`?:?`, ?, needs-coordination)
  - Emitting two rel=search links with a 'method' field requires adding a Method property to the Link record in Honua.Protocols.Ogc.Shared, which is outside the allowed file scope.
- [ ] **odata-conformance-14** (`?:?`, ?, needs-coordination)
  - Mapping geo.length in the OData parser to GEOLENGTH requires a corresponding case in PostgresSqlFilterTranslator (src/Honua.Postgres/) which is outside src/Honua.Geometry/.
- [ ] **cloud-io-host-7** (`?:?`, ?, needs-coordination)
  - Adding ControlPlaneTelemetry.Activities.BackendPromote requires editing src/Honua.Jobs/Features/ControlPlane/ControlPlaneTelemetry.cs, which is outside the src/Honua.Aws/ boundary.
- [ ] **spec-di-lifetimes-0** (`?:?`, ?, needs-coordination)
  - Changing Program.cs:447 to TryAddSingleton<IRoleStore> would activate the captive-dependency in OperatorAuthorizationEvaluator (singleton in src/Honua.Hosting/Features/Authentication/OperatorAuthorizationEvaluator.cs, outside src/Honua.Server/) which consumes IRoleStore; GeoprocessingJobService (singleton in src/Honua.Geoprocessing/) also transitively depends on IOperatorAuthorizationEvaluator, so resolving the full chain requires coordinated changes across Honua.Hosting and Honua.Geoprocessing that are outside the allowed edit scope.

## Acceptance criteria
- Each checkbox above either lands in a dedicated PR or is explicitly declined with a comment.
- Cross-project API changes follow the proto/SDK ownership rules where applicable.

## Affected repos
honua-server (some items may touch geospatial-grpc / SDKs when wire contracts move).

## Gate impact
PR gates only; no nightly/release gate changes.

## Non-goals
Re-running the sweep itself; in-project fixes already landed in #1563.

## Additional items (post re-grade)

- [x] **Lost-update window in UpdateProgressAsync read-modify-write** (`src/Honua.Core/Features/Temporal/Services/InProcessTemporalCorrectiveJobSink.cs:47`, medium, contract)
  - Correct fix is compare-and-set semantics on IUniversalProgressStore (cross-project interface); excluded from the sweep PR as a contract change.
- [x] **Push replica-sync ID filter into IChangeTracker/SQL** (`src/Honua.Core/Features/FeatureStore/Services/ReplicaSyncService.cs:113`, medium, needs-coordination)
  - The sweep bounded the conflict map by upload size, but a long-offline replica still materializes the full server change log; pushing the uploaded-ID filter into IChangeTracker (provider-side SQL) is the real fix.

mikemcdougall:
# Pull Request

## Issue Link
Fixes #1593

## Summary
The coordinated quality-sweep follow-ups **#1567** and **#1599** were merged into the `chore/assembly-grade-sweep` branch *after* that branch had already squash-merged to `trunk` (#1563 / `ab447ec`). Their commits were orphaned on the now-stale sweep branch (`f031046`, 12 commits ahead of trunk) and **never reached `trunk`** — despite #1593's status comment listing them as landed. Verified by absence on `origin/trunk`: `ProgressCompareAndSetResult.cs`, `FeatureEditPrecondition.cs`, `VersionExceptions.cs`.

This PR re-applies the three orphaned commits onto current `trunk` and reconciles them with 24 commits of independent drift — most of which applied cleanly; the OData layer required per-behavior conformance decisions (recorded on #1593).

## Changes Made
- Cherry-picked the 3 orphaned commits onto current `trunk` (#1567 ×2, #1599); resolved conflicts by keeping trunk's already-superseding work (durable-RBAC scope factory #1575, `FeatureServerEditsKey` #1614) and re-applying the dropped `RedisVersionJobStore` logger arg.
- **OData reconciliation** per #1593 decisions: keep trunk's reject→400 for absolute batch sub-request URLs (revert #1567 §11.7); keep trunk's `$top`=page-size paging (revert #1599 §11.2.6.7 `clientTop` short-circuit; restore `Transfer-Encoding: chunked`); **keep** the #1567/#1599 `$count`-on-empty-page fix (§11.2.5.5) and correct trunk's test that asserted the pre-fix bug (count 0 → 15).
- Kept non-conflicting recovered OData work: filter parser (decimal literals, `geo.length`, null-safe `ne`/`eq`), batch TOCTOU edits, CRUD/edit adapters, `$orderby` null placement, string-eq case sensitivity.

## Testing
- [ ] Architecture tests pass

## Gate Impact
- [ ] None — no gate impact

## Docs or Contract Impact
- [ ] OpenAPI spec changed
- [ ] Documentation updated

## Release/Deploy Impact

## Breaking Changes
None. Two OData v4.01 conformance behaviors (§11.7 absolute batch URLs, §11.2.6.7 `$top`) were intentionally *not* adopted to preserve trunk's current client-visible behavior; they are re-filed as decisions on #1593.

---

## Pre-PR Checklist


mikemcdougall:
Automated triage: this PR is **not mergeable as-is** — ~81 deterministic test failures across 6 subsystems (build passes; the shared test fixture is fine). Root causes are source/test reconciliation conflicts + genuine regressions introduced by the recovery merge. Specific items for the author:

1. **OGC Maps conformance URIs (10 failures):** `OgcMapsConformanceHandler.cs` was reverted from `https://…/conf/*` back to `http://…/conf/*`, but trunk already settled on `https://` (and trunk tests assert it). Drop the revert (keep `https://`), or justify a re-revert and update the trunk tests in the same commit.
2. **License gate (Migration shard, 37 failures):** the new `LicenseGate.RequireEntitlement(…, "import.geoservices"/"import.geoserver")` on the GeoServices/GeoServer import discover+start endpoints returns **402** where tests expect 400/success. Grant those entitlements in the integration-test license profile, or move the gate so discovery/validation still 400s on bad input; update tests where 402 is intended.
3. **PBF M-ordinate regression (real bug):** the rewritten `PbfQueryFormatter.cs` flips the M-ordinate sign (`2500000` → `-2500000`). Fix the delta/zigzag M encoding.
4. **SQL translator geometry cast (Postgres Compatibility):** generic `ST_LENGTH`/`ST_AREA` emit `ST_Length("geom")` instead of `ST_Length("geom"::geometry)`; route geometry props through the geometry-column expression (or adjust the new `KeepsPlanarLength` tests).
5. **FileImport count regression:** continue-on-error CSV import returns `featureCount:0` where 3 rows were imported — restore feature counting on that path.

Re-trigger the gated CI matrix with a push (not a label toggle) after fixing. No changes were pushed by triage.

PR #1722, opened 2026-06-17
test(cesium): pin transitive dompurify to 3.4.1 (fix js-cesium WMS render cert)

mikemcdougall:
## Root cause
The **Real-Client Interop Matrix (Nightly)** release gate regressed (runs [27616613043] x2, 2026-06-16) with exactly one cell: `js-cesium/wms: CERT-RNDR-01 pass → fail` ("Map/table renders without client error"). Confirmed **deterministic** (reproduced on re-run), **not** a flake.

Traced to dependabot **#1705**, which floated Cesium's **transitive** `dompurify` dependency (`cesium@1.119.0` requires `^3.3.0`) from **3.4.1 → 3.4.10** via a lockfile-only change. dompurify 3.4.10 is an internal refactor with no documented public API break, but Cesium 1.119.0's internal sanitize path errors against it during WMS imagery rendering → the browser client error CERT-RNDR-01 asserts on.

`dompurify` is **not imported by this harness and is not a shipped product dependency** — it's a transitive dep of a test-only browser fixture, so there is no product-security impact to pinning it.

## Fix
Add `overrides: { "dompurify": "3.4.1" }` to `tests/js-browser/cesium/package.json` and regenerate the lockfile, holding dompurify at the last-green version. Lock now resolves `@cesium/engine/node_modules/dompurify@3.4.1`.

## Validation
- Lockfile resolves dompurify → 3.4.1 (was 3.4.10). ✅
- Full validation is the **interop nightly going green** on this branch / after merge (the Playwright WMS spec requires a live server backend, not run locally).

## Follow-up
Tracked separately: re-bump dompurify properly (either to a 3.4.x that Cesium 1.119 tolerates, or bump Cesium) instead of pinning indefinitely.

### DIFF
--- a/tests/js-browser/cesium/package-lock.json
+++ b/tests/js-browser/cesium/package-lock.json
@@ -51,6 +51,15 @@
         "node": ">=20.19.0"
       }
     },
+    "node_modules/@cesium/engine/node_modules/dompurify": {
+      "version": "3.4.1",
+      "resolved": "https://registry.npmjs.org/dompurify/-/dompurify-3.4.1.tgz",
+      "integrity": "sha512-JahakDAIg1gyOm7dlgWSDjV4n7Ip2PKR55NIT6jrMfIgLFgWo81vdr1/QGqWtFNRqXP9UV71oVePtjqS2ebnPw==",
+      "license": "(MPL-2.0 OR Apache-2.0)",
+      "optionalDependencies": {
+        "@types/trusted-types": "^2.0.7"
+      }
+    },
     "node_modules/@cesium/wasm-splats": {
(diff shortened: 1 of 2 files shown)


mikemcdougall:
✅ **Validated.** Full client-interop matrix dispatched on this branch (run 27658369858) is **green** — `js-cesium/wms CERT-RNDR-01` passes with the dompurify 3.4.1 pin.

Note: validating with a single-lane dispatch (`lanes=cesium`) gives a false regression — a partial run emits 0 current envelopes, so the baseline-diff flags every cell as missing. Validate with the **full matrix** (default lanes). Follow-up to un-pin tracked in #1723.

PR #1969, opened 2026-06-21
test: isolate DbUp migration tests in a dedicated database

mikemcdougall:
## Problem

`DbUpMigrations_MobileOfflineDemoSeed_AppliesToCanonicalSchema` (in `DatabaseMigrationTests`) is a concurrency/isolation flake that taxes the serial merge-train with retries (it goes green on re-run).

Root cause: the test runs raw DbUp migrations and the `mobile-offline-demo` seeds, all of which target the **hard-coded `honua`/`honua_data` schemas**. The `Database.*` collections run in parallel against a single shared PostGIS database and rely on per-test `search_path` schema isolation — which does **not** scope fixed-name schemas. So the test's `COUNT(*)` assertions on `honua.services` / `honua.layers` intermittently collide with sibling collections that also create/populate the process-global `honua` schema.

## Fix

Give the migration-test class its **own uniquely-named database** so its `honua` schema, catalog DDL locks, and seed data are fully private and cannot collide with parallel collections:

- New `PostgresFixture.CreateIsolatedDatabaseAsync` / `DropDatabaseAsync` — create a per-test database with the standard Honua spatial extensions enabled, and tear it down (terminating lingering backends first).
- `DatabaseMigrationTests` now provisions a dedicated database in `InitializeAsync` and runs all migrations/seeds/assertions against it. A per-test schema inside that database preserves the `current_schema()` storage-binding semantics the seed asserts. With a dedicated database the prior advisory-lock serialization workaround is unnecessary on this path.

This is the database-per-collection hardening previously called out as the durable fix for these tests.

## Verification

- Release build, warnings-as-errors: clean (0 warnings, 0 errors).
- `dotnet format --verify-no-changes` on the changed files: clean.
- `DatabaseMigrationTests` (5 tests): **6 consecutive serial runs, 5/5 passing each**.
- Seed-consuming sibling `MobileOfflineDemoFixtureReplicationTests` (GeoServices): **7/7 passing**, confirming the shared-schema siblings are unaffected.


mikemcdougall:

PR #2002, opened 2026-06-21
fix(protocols): OGC bbox/datetime, GeoServices envelope, WMS TIME, OData skiptoken conformance (#1987,#1988,#1990,#1991)

mikemcdougall:
## Issue Link
Closes #1987
Closes #1988
Closes #1990
Closes #1991
Related to #1989

## Summary
A batch of protocol-conformance fixes:

- **OGC API Features + Records (#1987):** a 6-element (3D) bbox `minx,miny,minz,maxx,maxy,maxz` is spec-legal; instead of 400-ing 3D-aware OGC/ArcGIS clients, use the horizontal coordinates and ignore Z.
- **GeoServices (#1990):** require exactly four coordinates for the `esriGeometryEnvelope` branch. A short list (e.g. `1,2`) previously read xmax/ymax from the zero-initialized `stackalloc` tail and silently queried a different window instead of returning 400.
- **WMS GetFeatureInfo (#1991):** parse and apply the TIME dimension, mirroring GetMap, so identify honors the selected instant on time-enabled layers and an invalid TIME yields `InvalidDimensionValue`.
- **OGC Records / STAC (#1988):** stop dropping records that have no temporal value when a `datetime` filter is present (which made every datetime query return zero results); reject `limit=0` with 400 on the STAC items surface to match POST `/search`. The `/queryables` routes and the datetime-syntax-vs-no-temporal-field split are already resolved on `trunk`.
- **OData (#1989, partial):** a bare-integer `$skiptoken` now requires no `$filter`/`$orderby`; accepting it alongside a filter let a token minted for one query be reused as a raw offset against a different one, defeating the query-fingerprint guard. The remaining parts of #1989 (`$top=0`, Layers `@odata.nextLink`, provider `TotalCount` trust) are higher-risk query-path changes left for a follow-up, so this stays `Related to #1989`.

## Changes Made
- `OgcFilterProcessor` + `OgcRecordsEndpoints`: accept 4- or 6-element bbox.
- `GeoServicesGeometryParser`: require `coordinateCount == 4` for envelope; 400 otherwise. + unit test.
- `WmsRequestHandlers.GetFeatureInfo`: parse TIME and thread `TemporalFilter` per query layer.
- `OgcRecordsEndpoints`: do not drop temporally-null records under a datetime filter.
- `Stac/ItemEndpoints`: 400 on `limit < 1`.
- `ODataSkipTokenService`: reject bare-integer token when a filter/orderby is present. + integration test.

## Testing
[...]

PR #2030, opened 2026-06-22
fix(odata): v4 conformance — $top=0, Layers nextLink, robust paging (#1989)

### LINKED ISSUE #1989: OData v4 conformance: $top=0 → 400, missing nextLink, skiptoken fingerprint bypass
**Severity:** Medium · **Confidence:** Medium-High

1. **`$top=0` → HTTP 400** (v4 violation) — `src/Honua.Protocols.OData/Features/ODataValidationService.cs:17-18` uses `MinLimit: 1`. OData v4 allows `$top=0` (empty `value`, often with `$count=true`). SDKs that do this get a 400.
2. **Layers collection emits no `@odata.nextLink`** — `Features/ODataQueryHandler.cs:48-173` paginates in memory but issues no `$skiptoken`; `/odata/Layers?$top=10` over >10 layers returns 10 with no continuation → clients conclude there are no more.
3. **Legacy integer `$skiptoken` bypasses the query fingerprint** — `Features/Services/ODataSkipTokenService.cs:60-71`: opaque tokens are bound to a SHA-256 fingerprint of filter+orderby, but a bare-integer `$skiptoken=500` is accepted as a raw offset with no fingerprint check, defeating the cross-query-reuse guard.
4. **Non-streaming pagination trusts provider `TotalCount`** — `Features/Services/ODataUtilityService.cs:447-450`: `ShouldPaginate` ignores its `top` param and computes `currentSkip + resultCount < totalCount`. Correct only because the Postgres provider emits `COUNT(*) OVER ()`; if `TotalCount` degrades to `features.Count` (the `ExtractTotalCount` fallback, or a non-Postgres read-only provider), nextLink is suppressed and paging silently truncates. The streaming path guards this with a +1-row probe; this path doesn't.

---
*Filed from an automated org-wide bug-bounty code review. Each finding was read at the cited `file:line`. Confidence noted above; verify before fixing.*

mikemcdougall:
Closes #1989

Remaining OData v4 parts (skiptoken-fingerprint part already in #2002):
- `$top=0` returns 200 empty with correct `$count` (negative still 400)
- Layers collection emits a followable `@odata.nextLink`
- non-streaming pagination uses a +1-row continuation probe instead of trusting provider `TotalCount`

Does not touch `ODataSkipTokenService.cs`. Builds clean (Release, TWAE); 50/50 unit tests pass; integration tests need Docker/PostGIS in CI.


mikemcdougall:

mikemcdougall:

mikemcdougall:

PR #2072, opened 2026-06-24
fix(security): RLS edit/cache bypass, replica watermark, version-post lost-update, densify/exportTiles DoS

### LINKED ISSUE #2062: Replica sync watermark uses sequence last_value → concurrent committed edits silently dropped from replicas (and clobbered on upload)
**Severity:** High · **Confidence:** High · silent data loss across replicas

**Locations:**
- Cursor read: `src/Honua.Postgres/Features/FeatureStore/Services/PostgresChangeTracker.cs:27` — `SELECT last_value FROM honua.sync_generation`
- Generation assignment: `src/Honua.Server/Migrations/012_AddReplicationDurability.sql:10` (`CREATE SEQUENCE honua.sync_generation`), trigger `050_AddChangeLogAttribution.sql:65` — `gen := nextval('honua.sync_generation')` inside each edit's own transaction
- Cursor persisted from `last_value`: `src/Honua.Protocols.GeoServices/FeatureServer/FeatureServerRequestHandlers.Replication.cs:889-896` (`currentGen = GetCurrentGenerationAsync(); LastSyncGeneration = currentGen`); also :251,:348,:477
- Next-delta strict bound: `PostgresChangeTracker.cs:61` — `WHERE generation > $1`

**What's wrong:** A Postgres sequence is **non-transactional**: `nextval` advances `last_value` globally and immediately and is never rolled back, while the corresponding `feature_changes` row only becomes visible when *that* transaction commits. Concurrent edit transactions commit out of generation order. `GetCurrentGenerationAsync` reads `last_value` (the max consumed by *any* session, including in-flight transactions) and stores it as the replica's `LastSyncGeneration`. The next delta selects strictly `generation > cursor`, so any change whose generation `<= cursor` that commits *after* the cursor was read is skipped forever.

**Concrete data-loss sequence:**
1. Tx A inserts feature X; trigger consumes `nextval` = 100; A not yet committed.
2. Replica R syncs; `GetCurrentGenerationAsync` reads `last_value` = 100; R persists `LastSyncGeneration = 100`.
3. Tx A commits; row `(gen=100, X)` becomes visible.
4. R's next sync queries `generation > 100` — X is never returned. Insert lost on the replica permanently (updates → stale row; deletes → replica keeps a server-deleted row).

**Secondary (same root cause):** On upload/apply, conflict detection uses `GetChangesSinceAsync(replica.LastSyncGeneration, …)` (`Replication.cs:853`) with hard last-write-wins (`:855`). A server edit committed in the same out-of-order window carries `generation <= BaseGeneration`, is invisible to conflict detection, and the replica's edit silently clobbers the committed server edit with no conflict recorded.

**Why certain:** Watermark is `last_value`; generations are assigned by an immediately-incrementing sequence inside separately-committing ReadCommitted transactions; next read uses strict `>`. The skip is mechanical under any concurrent editing. No xmin/snapshot guard or "advance only to max-observed-committed" logic exists.

**Fix:** Don't use `last_value` as the watermark. Either (a) gate the cursor by the oldest in-flight xmin (only advance past generations known fully committed, e.g. track `txid_snapshot`/`pg_snapshot` and exclude `xip` gaps), or (b) assign generations at commit time, or (c) make the delta query return rows by commit-visibility (a monotonic commit LSN / `pg_current_snapshot`) rather than a strict sequence cutoff.

---
*Filed from a strict second-pass write-path data-integrity audit.*

### LINKED ISSUE #2063: Branch-version POST overwrites concurrent committed DEFAULT edits (lost update; base_* columns never checked)
**Severity:** High · **Confidence:** High · lost update of committed DEFAULT edits

**Location:** `src/Honua.Postgres/Features/FeatureStore/Services/PostgresVersionManager.ReconcilePost.cs`
- `PostAsync` gate: lines 216-223 — proceeds whenever `CountPendingConflictsAsync(versionId) == 0`
- `ReplayOverlayOntoDefaultAsync`: UPDATE `:275-284` (`UPDATE features f SET geometry = ve.geometry, attributes = ve.attributes … WHERE f.layer_id = ve.layer_id AND f.objectid = ve.objectid`) and DELETE `:264-272` — keyed **only** on `(layer_id, objectid)`
- Conflict detection lives only in the separate `ReconcileAsync` (`:65-136`), which diffs `feature_changes` since `CommonAncestorGeneration`. The overlay carries `base_geometry`/`base_attributes` (`049_CreateVersionEdits.sql`), but POST never compares them to the live DEFAULT row.

**What's wrong:** POST trusts a *prior, separately-invoked* reconcile's persisted conflict set instead of re-checking DEFAULT at apply time. The version lock (`PostgresVersionManager.cs:27-33`) serializes only same-version reconcile/post — it does not block DEFAULT writers. So any DEFAULT edit committed between a clean reconcile and the post is undetected, and the whole-row UPDATE silently discards it.

**Concrete data-loss sequence:**
1. Branch V updates `(L,O)`; overlay base captured as `{name:"A", pop:100}`.
2. `ReconcileAsync(V)` runs clean → `pending = 0`, `common_ancestor_gen` advanced.
3. A DEFAULT editor commits UPDATE `(L,O)` → `pop:999` (committed to `features`).
4. `PostAsync(V)` runs; `pending` still 0; the UPDATE overwrites the entire DEFAULT row with the branch image, discarding the committed `pop:999` with no conflict raised.

**Why certain:** The POST UPDATE/DELETE SQL is keyed solely on `(layer_id, objectid)` with no `base_*`-vs-live comparison; `PostAsync`'s only guard is the stale `version_conflicts` count from a distinct operation; DEFAULT is never locked or re-diffed at post time. The columns needed to detect the conflict exist and are simply not read on this path.

**Fix:** At post time, re-validate each overlay row's `base_geometry`/`base_attributes` against the current DEFAULT row (optimistic check in the UPDATE `WHERE`, or a fresh reconcile under a lock that also blocks/▶snapshots DEFAULT), and raise a conflict instead of overwriting when they differ.

---
*Filed from a strict second-pass write-path data-integrity audit.*

### LINKED ISSUE #2064: GeometryServer /densify: unbounded vertex generation from maxSegmentLength (unauth OOM DoS)
**Severity:** High (unauthenticated OOM DoS) · **Confidence:** High · verified on `origin/trunk` (045da76b2)

**Location:** `src/Honua.Protocols.GeoServices/GeometryService/Services/GeometryServiceHandler.cs:2561`

**Unbounded expression:** `Densifier.Densify(geometry, parameters.MaxSegmentLength)` — NTS generates `ceil(segmentLength / maxSegmentLength)` interpolated `Coordinate`s per segment into an in-memory `CoordinateList`.

**Trace:**
- `POST /rest/services/.../GeometryServer/densify` is `.AllowAnonymous()` (`GeometryServiceEndpoints.cs:171-175`) — no auth.
- `maxSegmentLength` is parsed by `ParseRequiredPositiveDouble` (`:933-940`, impl `:2646-2660`), which rejects only `<= 0`/`NaN`/`Infinity`. **No lower bound and no cap relative to geometry extent** between parse and the `Densify` call. The geodesic branch (`:951-960`) only multiplies, never floors.
- A single polyline under the 10 MB per-geometry cap suffices.

**Trigger:**
```
POST /rest/services/Utilities/Geometry/GeometryServer/densify
f=json&sr=4326&geometries={"geometryType":"esriGeometryPolyline","geometries":[{"paths":[[[0,0],[2000000,0]]]}]}&maxSegmentLength=0.001
```
2,000,000 / 0.001 = 2e9 coordinates generated synchronously before any response → OOM / GC thrash / process wedge. The `catch (Exception)` at `:983` cannot recover from OOM-class host pressure.

**Fix:** Reject when `geometryExtent / maxSegmentLength` exceeds a bound, or enforce a minimum `maxSegmentLength` relative to geometry extent (cap total output vertices).

---
*Filed from a strict second-pass crash/DoS audit; verified on origin/trunk.*

### LINKED ISSUE #2065: MapServer exportTiles/estimateExportTilesSize materializes full tile grid before maxTiles cap (unauth OOM DoS)
**Severity:** High (unauthenticated OOM DoS) · **Confidence:** High · verified on `origin/trunk` (045da76b2)

**Location:** `src/Honua.Protocols.GeoServices/MapServer/MapServerRequestHandlers.ExportTiles.cs:368` (grid built at `:786-812`)

**Unbounded expression:** `var allTiles = BuildExportTileCoordinates(bounds, requestedZooms);` — nested loop (`:802-808`) `for x in [xMin..xMax] for y in [yMin..yMax] coordinates.Add(...)` over the full grid; `.Take(maxTiles)` is applied **after** at `:369-370`, too late to bound the allocation.

**Trace:**
- `estimateExportTilesSize` (GET+POST) is `.AllowAnonymous()` (`MapServerEndpoints.cs:79-84`, anon at `:60`), and `BuildExportTileCoordinates` runs at `:368` **before** `ValidateServiceV2Async` (`:377`) and the access-policy check (`:393`) — reachable on a non-existent service with no auth.
- Omitting `exportExtent` defaults to the whole world (`TryParseExportTilesExtent:647-651`, `-180..180` × full WebMercator lat).
- `maxZoom` is validated only per-level against `[MinTileZoom, MaxTileZoom]`; `MaxTileZoom = 18` (`src/Honua.Core/Configuration/TileLimits.cs`). No cap on per-level tile count or bbox-area × zoom product.
- At z18 over default world extent: `n = 2^18 = 262144`, full span → `n × n ≈ 6.9e10` `ExportTileCoordinate` entries in one `List` (hundreds of GB) → OOM. `maxTiles` (clamped to `MaxTilesPerRequest` at `:636`) only bounds the post-hoc `.Take`, not the build.

**Trigger:**
```
GET /rest/services/anything/MapServer/estimateExportTilesSize?maxZoom=18&maxTiles=1
```

**Fix:** Compute the tile count `Σ (xMax-xMin+1)*(yMax-yMin+1)` first and reject when it exceeds `maxTiles`/`MaxTilesPerRequest`, or stream the enumeration and stop after `maxTiles`. Also run validation/auth before building the grid.

---
*Filed from a strict second-pass crash/DoS audit; verified on origin/trunk.*

### LINKED ISSUE #2066: RLS bypass: GeoServices updateFeatures/deleteFeatures/applyEdits skip not-found guard on default-OBJECTID layers (edit/delete RLS-hidden rows)
**Severity:** High (RLS authorization bypass — unauthorized mutation/deletion) · **Confidence:** High · verified on `origin/trunk` (045da76b2)

**Location:** `src/Honua.Protocols.GeoServices/FeatureServer/FeatureServerEditsHandler.cs:382` (update) and `:474` (delete)

**The wrong check:** the "feature not found" guard is conditional:
```csharp
if (!ShouldUseInternalObjectIdFastPath(resource) && existingFeature is null) { /* reject */ }
var internalObjectId = existingFeature?.Id ?? objectId;   // :392 / :484
```
`ShouldUseInternalObjectIdFastPath` (`:1109-1112`) returns `true` whenever the layer's objectId field is the default `OBJECTID` (the common case). On that path the null-`existingFeature` rejection is **skipped**, and `internalObjectId` falls back to the **client-supplied** `objectId`.

**Why it's a bypass:** the pre-edit resolver `ResolveFeaturesByGeoServicesObjectIdsAsync` (`:1051-1067`) IS RLS-enforced (reads via `_featureReader.QueryAsync` → `PostgresFeatureStore.ExecuteFormatQueryAsync` → `ApplyPermanentFilterAsync` → `ResolveRlsFilterAsync`, `PostgresFeatureStore.cs:653/903/948`), so an RLS-hidden row yields `existingFeature == null`. But with the guard skipped, the hidden objectId is still queued, and the edit SQL applies **no** RLS — `ApplyEditsAsync` is a pass-through (`PostgresFeatureStore.cs:552-555`) and the UPDATE/DELETE filters only `WHERE layer_id = $1 AND objectid = $2` (`FeatureDataAccess.Edits.cs:749,804,827`).

**Exploit path:** An authenticated low-privilege user, on an RLS-protected layer with the default `OBJECTID` field, issues `applyEdits`/`updateFeatures`/`deleteFeatures` with the objectId of a row RLS hides from them. The row is deleted, or its attributes/geometry overwritten — a row the caller may not see or touch. The custom-objectid-field path (guard active) correctly returns "Feature not found", and the `deleteFeatures?where=…` path resolves IDs via the RLS-enforced query first, so both are safe; the gap is specific to explicit-objectId edits on default-OBJECTID layers. Replica-sync edits share `HandleApplyEditsAsync` and inherit the gap.

**Why certain:** All four legs read — the conditional guard, the RLS-enforced resolver, the non-RLS `ApplyEditsAsync` pass-through, and the objectid-only edit SQL.

**Fix:** Make the not-found rejection unconditional (reject when the RLS-enforced pre-read returns null regardless of fast-path), or apply the RLS predicate to the edit SQL's WHERE clause.

---
*Filed from a strict second-pass auth/RBAC audit; verified on origin/trunk.*

### LINKED ISSUE #2067: Cross-principal RLS leak: MapServer cloud tile cache key omits principal/RLS dimension
**Severity:** High (cross-principal data leak) · **Confidence:** High · verified on `origin/trunk` (045da76b2) · precondition: cloud tile storage configured + enabled

**Location:** `src/Honua.Protocols.GeoServices/MapServer/MapServerRequestHandlers.Tile.cs:249-288` (`BuildMapServerTileCacheKey`), consumed at `:126` (read) and `:154` (write)

**The missing dimension:** the cache key hashes only `snapshot.Etag | service.Metadata.Id | serviceSrid | maxFeatures | renderLayerKey` + tile coords (`:270-287`) — **no principal / role / claim / RLS dimension**. Unlike the OGC and VectorTileServer output caches (anonymous-only policy, skip caching authenticated requests), `GeoServicesCloudTileCache` is a separate persistent cloud-storage cache with no such guard — active whenever cloud `storage` is registered and `storageOptions.Enabled != false` (`GeoServicesCloudTileCache.cs:30,83`).

The rendered PNG genuinely contains RLS-filtered feature geometry: `RenderRasterTileV2Async` → `RasterMapRenderingPipeline` fetches via the request-scoped `IFeatureReader.QueryAsync`, which applies the caller's RLS predicate.

**Exploit path:** User A (broad RLS claims) requests a MapServer tile; the PNG rendered from A's RLS-visible rows is written under the principal-agnostic key. User B (narrower RLS claims) requests the same service/z/x/y, passes only the coarse service-level `RequireAnyResourceAccess` gate (`Tile.cs:84`), and is served A's cached tile (`:126-132`) — seeing geometry RLS should hide from B. (The ImageServer tile cache uses the same key but serves raster-mosaic imagery, not RLS feature rows — lower impact.)

**Why certain:** Read the key builder (no principal input), the unconditional read/write around it, the cache-enable gate, and confirmed the render path uses the request-scoped RLS reader. Distinct cache and data class from the already-filed metadata-output-cache license-tier bleed.

**Fix:** Add a principal/RLS fingerprint (effective RLS predicate hash, or tenant+role+claims) to the tile cache key, or disable the cloud tile cache for RLS-governed layers / authenticated principals.

---
*Filed from a strict second-pass auth/RBAC audit; verified on origin/trunk.*

mikemcdougall:
# Pull Request

## Issue Link
Closes #2066
Closes #2067
Closes #2062
Closes #2063
Closes #2064
Closes #2065

## Summary
Fixes six verified P1 security/integrity bugs found in the second-pass audits, one focused commit each (plus three test commits). Two RLS authorization bypasses, two write-path data-loss bugs, and two unauthenticated OOM DoS vectors. They span Protocols.GeoServices, Postgres, and Server migrations — fine for one PR since each is independently scoped.

## Changes Made
- **#2066 (RLS edit bypass):** `FeatureServerEditsHandler` skipped the not-found rejection on the default-`OBJECTID` fast path, letting a caller update/delete an RLS-hidden row by its OBJECTID. The RLS-enforced pre-read returns null for a hidden row, but the edit SQL filters only on `(layer_id, objectid)`. Made the not-found rejection **unconditional** for both update and delete. The custom-objectid and `?where=` paths were already safe.
- **#2067 (cross-principal RLS tile-cache leak):** `BuildMapServerTileCacheKey` omitted any principal/RLS dimension, so principal A's RLS-filtered PNG could be served to principal B for the same z/x/y. Resolve the effective RLS predicate (SQL + bound claim values) for the tile's render layers via `IRowLevelSecurityFilterSource` and fold a stable hash into the cache key. ImageServer's tile cache is raster-mosaic only (no RLS feature reader) — left unchanged, matching the issue's assessment.
- **#2062 (replica-sync watermark data loss):** The cursor read the non-transactional sequence `last_value`, so a lower-generation edit committing out of order after the watermark was read was skipped on replicas forever. Switched `PostgresChangeTracker.GetCurrentGenerationAsync` to `COALESCE(MAX(generation),0) FROM honua.feature_changes`, and added a generation-allocation advisory lock to the DEFAULT and version change-tracking triggers (migration `067`) so generation order equals commit order — mirroring the proven `fieldcollection_changes` pattern.
- **#2063 (branch-post lost update):** `ReplayOverlayOntoDefaultAsync` keyed UPDATE/DELETE only on `(layer_id, objectid)`, gated solely by a stale prior-reconcile conflict count, silently overwriting DEFAULT edits committed since. Now re-validates each update/delete overlay row's captured `base_geometry`/`base_attributes` against the live DEFAULT row inside the post transaction (`IS NOT DISTINCT FROM`, NULL-safe; matched rows pinned `FOR UPDATE`); on any divergence it rolls back and surfaces a conflict instead of clobbering. The optimistic predicate is also applied to the UPDATE/DELETE WHERE as defense-in-depth.
- **#2064 (densify OOM DoS):** Anonymous `/densify` passed `maxSegmentLength` straight to NTS, allowing ~2e9 interpolated vertices from a tiny value over a large extent. Estimate the output vertex count (`geometry.Length / maxSegmentLength`) and reject over a per-geometry cap before densifying.
- **#2065 (exportTiles OOM DoS):** `estimateExportTilesSize`/`exportTiles` built the full tile grid (z18 whole-world ≈ 6.9e10 entries) before `.Take(maxTiles)`. Now compute the count `Σ(xMax-xMin+1)*(yMax-yMin+1)` first (long arithmetic), reject/flag when it exceeds maxTiles, run service validation and the access-policy gate before building the grid, and bound the build to maxTiles coordinates. The estimate response reports the true total count with `ExceededTransferLimit`.

## Testing
- [x] Architecture tests pass

Added focused tests: RLS edit/delete not-found guard on a default-OBJECTID layer (#2066); densify vertex-cap rejects the DoS payload while a reasonable request still succeeds (#2064); exportTiles whole-world high-zoom estimate returns promptly with the true count + `ExceededTransferLimit` instead of OOM (#2065). Full solution builds clean under Release + `TreatWarningsAsErrors` (0 warnings, 0 errors); `dotnet format --verify-no-changes` clean on all changed files; change-tracker unit tests pass. Docker was unavailable locally, so the new Testcontainers-backed integration tests run in CI. Two pre-existing `feature-catalog.json` drift failures around `oauth-clients`/`oidc-providers` admin endpoints are unrelated to this PR (none of these changes touch those areas).

## Gate Impact

## Docs or Contract Impact

## Release/Deploy Impact
- [x] Requires database migration

Adds migration `067_SerializeChangeLogGenerationAllocation.sql` (`CREATE OR REPLACE` of two existing trigger functions to take a generation-allocation advisory lock before `nextval`). Idempotent and backward-compatible; every other line is byte-identical to migration `050`.

## Breaking Changes
None

---

## Pre-PR Checklist
- [x] If protocol/auth behavior changed: updated compatibility contract

### DIFF
--- a/src/Honua.Postgres/Features/FeatureStore/Services/PostgresChangeTracker.cs
+++ b/src/Honua.Postgres/Features/FeatureStore/Services/PostgresChangeTracker.cs
@@ -24,7 +24,18 @@ internal sealed class PostgresChangeTracker : IChangeTracker
 
     public async Task<long> GetCurrentGenerationAsync(CancellationToken cancellationToken = default)
     {
-        const string sql = "SELECT last_value FROM honua.sync_generation";
+        // Committed high-watermark for replica sync. The honua.sync_generation sequence reports
+        // its last ALLOCATED value (last_value), which includes generations consumed by edit
+        // transactions that have called nextval but not yet committed. A replica that persisted
+        // last_value as its cursor and then selected strictly generation > cursor would skip any
+        // lower-generation change that committed after the cursor was read (commit order != gen
+        // order under concurrency), losing it from the replica forever (#2062). Reading the
+        // committed MAX(generation) from honua.feature_changes instead gives a watermark the
+        // puller can advance to without skipping an in-flight write; migration 067 takes a
+        // generation-allocation advisory lock in the change-tracking triggers so generation order
+        // equals commit order and this committed-MAX watermark is gap-free. Mirrors the proven
+        // honua.fieldcollection_changes pattern (PostgresFieldCollectionSyncStore).
+        const string sql = "SELECT COALESCE(MAX(generation), 0) FROM honua.feature_changes";
 
         await using var connection = await _connectionProvider.OpenNpgsqlConnectionAsync(cancellationToken).ConfigureAwait(false);
         await using var command = new NpgsqlCommand(sql, connection);
--- a/src/Honua.Postgres/Features/FeatureStore/Services/PostgresVersionManager.ReconcilePost.cs
+++ b/src/Honua.Postgres/Features/FeatureStore/Services/PostgresVersionManager.ReconcilePost.cs
@@ -229,13 +229,29 @@ internal sealed partial class PostgresVersionManager
         await SetVersionStateAsync(versionId, VersionState.Posting, cancellationToken).ConfigureAwait(false);
         try
         {
-            var (applied, serverGeneration) = await ReplayOverlayOntoDefaultAsync(versionId, cancellationToken).ConfigureAwait(false);
-            await AdvanceCommonAncestorAsync(versionId, serverGeneration, cancellationToken).ConfigureAwait(false);
-            activity?.SetTag("honua.version.applied_changes", applied);
-            activity?.SetTag("honua.version.server_generation", serverGeneration);
+            var replay = await ReplayOverlayOntoDefaultAsync(versionId, cancellationToken).ConfigureAwait(false);
+
+            // A clean CountPendingConflictsAsync only reflects the conflict set computed by the LAST
+            // reconcile. The version lock serializes reconcile/post for THIS version but does not block
+            // DEFAULT writers, so a DEFAULT edit committed between that reconcile and this post would be
+            // silently overwritten by the whole-row overlay replay (lost update). ReplayOverlayOntoDefault
+            // re-validates each overlay row's captured base_* image against the LIVE DEFAULT row inside the
+            // post transaction; if any update/delete target diverged, it rolls back and reports the
+            // divergent keys here so the caller must reconcile again instead of clobbering (#2063).
+            if (replay.Conflicted)
+            {
+                activity?.SetTag("honua.version.conflict_count", replay.ConflictingKeys.Count);
+                activity?.SetTag("honua.version.outcome", "blocked_by_default_drift");
+                activity?.SetStatus(ActivityStatusCode.Ok);
+                return new VersionPostResult(Posted: false, AppliedChanges: 0, ServerGeneration: 0, BlockedByConflicts: true);
+            }
+
+            await AdvanceCommonAncestorAsync(versionId, replay.ServerGeneration, cancellationToken).ConfigureAwait(false);
+            activity?.SetTag("honua.version.applied_changes", replay.Applied);
+            activity?.SetTag("honua.version.server_generation", replay.ServerGeneration);
             activity?.SetTag("honua.version.outcome", "posted");
             activity?.SetStatus(ActivityStatusCode.Ok);
-            return new VersionPostResult(Posted: true, AppliedChanges: applied, ServerGeneration: serverGeneration, BlockedByConflicts: false);
+            return new VersionPostResult(Posted: true, AppliedChanges: replay.Applied, ServerGeneration: replay.ServerGeneration, BlockedByConflicts: false);
         }
         finally
         {
@@ -251,7 +267,7 @@ internal sealed partial class PostgresVersionManager
     /// DEFAULT change-tracking trigger fires per row (no version GUC set), advancing the generation and
     /// recording NULL-version DEFAULT changes. Returns the applied row count and the post generation.
     /// </summary>
-    private async Task<(int Applied, long ServerGeneration)> ReplayOverlayOntoDefaultAsync(
+    private async Task<OverlayReplayResult> ReplayOverlayOntoDefaultAsync(
         Guid versionId,
         CancellationToken cancellationToken)
     {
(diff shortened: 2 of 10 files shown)


mikemcdougall:

mikemcdougall:

mikemcdougall:

mikemcdougall:

mikemcdougall:

PR #2200, opened 2026-06-26
feat(geoprocessing): GP Devkit native-profile container-exec fidelity (#2180)

### LINKED ISSUE #2180: GP Devkit — local-runner native-profile fidelity: run gdal.* against the real worker container
From the GP architecture review. The devkit's `GeoprocessingLocalRunner` (P2) bypasses Redis/control-plane for a sub-second loop — **fine for managed-profile ops** (same executor code runs locally and in prod). But for **`native` (gdal.*) ops**, the local path doesn't exercise the GDAL-worker IMAGE, the claim fence, or Batch container packaging — so a `gdal.hillshade` that passes `gp run` locally can still fail at the worker-image / Batch boundary it never touched.

**Harden the inner loop so it's a true dry-run of the submit path:**
- Run `native`-profile steps against the **real GDAL worker container** (docker), not just an in-process path — so the image/CRS/arg handling is actually exercised locally.
- Have `gp run`/`gp plan` emit the **identical `ExecutionJobSpec`** the prod submit path would produce, so 'plan' validates the real spec rather than a parallel code path.
- Keep the managed-profile fast path as-is (no regression to the sub-second loop for the common case).

Severity: medium — fidelity gap, not a correctness bug. The local loop is sound for managed ops today; this closes the native-op cliff. Relates to the GP Devkit (#2130) and the local-dev quickstart (#2131, which already brings up the real worker-gdal container).

mikemcdougall:
## Summary

GP Devkit native-profile fidelity (#2180). The P2 local runner runs the real executor in-process — full fidelity for **managed** ops, but for **native** (`gdal.*`) ops the in-process path never crosses the image / CRS-data / driver-set / arg-handling boundary a production native submit crosses (the job is packaged into the GDAL worker container and dispatched by Batch). A `gdal.hillshade` that passes `gp run` locally could still fail at that boundary.

This adds an **opt-in** container-exec mode that runs each native `gdal.*` step inside the **real `honua-worker-etl` image** via `docker run`, and makes `gp run`/`gp plan` emit the **identical `ExecutionJobSpec`** the prod submit path produces — so "plan" is a true dry-run of the real native spec, not a parallel representation. The fast in-process managed loop is unchanged.

**Stacks on the P2 devkit branch** (`feat/gp-devkit-p2-runner-cli`, unmerged) — base this PR against it / land after P2.

## Changes Made

- Add `DockerGdalCommandRunner` (opt-in `IGdalCommandRunner`): runs each GDAL tool inside the `honua-worker-etl` image via `docker run` instead of the host CLIs. Correct-by-construction — every executor does all file I/O in a per-job scratch workspace and passes absolute workspace paths to the tool, so the runner bind-mounts that workspace at the **identical absolute path** (`-v <ws>:<ws> -w <ws>`) and overrides the image entrypoint to the raw tool; the executor code path is byte-for-byte unchanged.
- Add `GdalProcessExecutorMode` (InProcess default / Container) on the `AddGdalProcessExecutors` seam; container mode swaps the runner but registers the identical executor set.
- Wire `honua gp run --real-worker` (alias `--container`, plus `--in-process`): auto-selects the container path only when the image is present locally (`GdalContainerProbe`) so the sub-second managed loop is never blocked on a pull.
- Split the docker launch behind `IDockerCommandInvoker` so the `docker run` argv (image ref, mount, working dir, user, entrypoint override, arg order) is unit-tested offline against a fake.
- Extract `GeoprocessingSpecBuilder` as the single source of truth for projecting an `AnalysisPlan` onto the durable `ExecutionJobSpec`. Both `GeoprocessingJobService.BuildSpec` (no-workload case) and the local runner build their spec through it from an equivalent single-step plan → identical spec (parameter bag, Kind/TargetKind/Backend/WorkloadName, data-driven RuntimeProfile: `native` for `gdal.*`, `null` for managed). Runner no longer emits the legacy `protocolProcessId` fallback.
- Classify the `Honua.Geoprocessing.Cli` dev tool as tooling in the module-dependency policy (it is a `dotnet tool`-packaged, AOT-excluded developer convenience the lean server never references) — also clears a pre-existing matrix violation.

## Breaking Changes

None. In-process is the default; container mode is opt-in. Managed-profile fast path unchanged.

## Docker correct-by-construction note

Docker is not available in the build env, so the container-exec path is made correct-by-construction and verified by the docker invocation it builds (`DockerGdalCommandRunnerTests`): `docker run --rm --network none --user 1001:1001 -v <ws>:<ws> -w <ws> --entrypoint <tool> honua-worker-etl <args…>`. The **end-to-end container run** is **CI / local-Docker-verified** — it needs a Docker daemon + the `honua-worker-etl` image (`docker build -f docker/worker-gdal/Dockerfile -t honua-worker-etl .`).

## Gate Impact

- [x] No gate impact (dev-tool / local-runner fidelity; lean serving image + AOT surface unaffected — `Honua.Server` does not reference the CLI)

## Testing

- [x] Ran the affected build (Release, warnings-as-errors) and unit tests; all green

`dotnet build` (Release, `TreatWarningsAsErrors`): Worker.Gdal + Geoprocessing + CLI — **0 warnings, 0 errors**. `dotnet format --verify-no-changes`: clean. Tests: **Worker.Gdal 10/10**, **Server.Tests 7/7** (3 new spec-equivalence + 4 existing local-runner), **Architecture 3/3** (ModuleDependencyPolicy + GeoprocessingIsolation). Container end-to-end run is CI/local-Docker-verified (no daemon in build env).

Closes #2180

### DIFF
--- a/docker/worker-gdal/README.md
+++ b/docker/worker-gdal/README.md
@@ -76,3 +76,45 @@ assembly (including `Honua.Server`), so the lean serving image's
 package-graph and cold-start budget are completely unaffected. The CLI binaries
 come from the base image layer. The `IGdalCommandRunner` seam also lets the
 executor logic be unit-tested without GDAL installed.
+
+## GP Devkit container-exec fidelity (`--real-worker`, #2180)
+
+The GP Devkit local runner (`honua gp run`) executes the real executor in-process
+for a sub-second loop. For **managed** ops that is full fidelity. For **native**
+(`gdal.*`) ops the in-process path never crosses the image / driver-set / CRS-data /
+arg-handling boundary a production native submit crosses (the job is packaged into
+*this* image and dispatched by Batch) — a `gdal.hillshade` that passes locally
+against a host's GDAL could still fail at that boundary.
+
+`honua gp run … --real-worker` (alias `--container`) closes that cliff: it runs each
+GDAL tool inside **this image** via `DockerGdalCommandRunner` instead of the host
+CLIs. It is **opt-in** and auto-selected only when the image is already present
+locally, so the managed loop is never blocked on a pull (`--in-process` forces the
+fast path). The runner is correct-by-construction:
+
+- Each native executor does all file I/O inside a per-job scratch workspace under
+  `GdalWorker:ScratchRoot` and passes absolute workspace paths to the tool.
+- `DockerGdalCommandRunner` bind-mounts that workspace at the **identical absolute
+  path** (`docker run --rm --network none --user 1001:1001 -v <ws>:<ws> -w <ws>
+  --entrypoint <tool> honua-worker-etl <args…>`), so the executor's absolute paths
+  resolve to the same files inside the container, and outputs land back on the host
+  workspace for read-back. The executor code path is byte-for-byte unchanged — only
+  the `IGdalCommandRunner` implementation differs.
+
+The `docker run` invocation (image ref, mount, working dir, user, entrypoint
+override, arg ordering) is unit-tested offline against a fake container-runtime seam
+(`DockerGdalCommandRunnerTests`); the **end-to-end container run** is
+CI / local-Docker-verified — it needs a Docker daemon and the `honua-worker-etl`
+image (`docker build -f docker/worker-gdal/Dockerfile -t honua-worker-etl .`).
+
+Build the image and run a native op against it:
+
+```sh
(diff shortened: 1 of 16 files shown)


mikemcdougall:
Landed on trunk via the Round-4 aggregate #2214 (ad0a0d150). Closing as merged-through; content was incorporated with conflict resolution so GitHub couldn't auto-detect the merge.

PR #2226, opened 2026-06-26
feat: durable Redis-backed stores for WFS stored queries, workflow packages, and role cache (#1593)

### LINKED ISSUE #1593: tech-debt: coordinated follow-ups from repo-wide quality sweep (cross-cutting + geodesy/standards reviews)

mikemcdougall:
## Issue Link
Closes #1593

## Summary
Adds durable, Redis-backed implementations for stores that were previously in-memory only, so WFS 2.0 stored queries, workflow packages, and role/permission lookups survive restarts and scale across instances. In-memory implementations are retained as the default fallback.

## Changes Made
- Add `IWfsStoredQueryStore` with in-memory and Redis implementations; refactor `Wfs20Handler` stored-query paths to use the store
- Add `RedisWorkflowPackageStore` alongside the in-memory store; wire registration
- Add `CachingRoleStore` and `RolePermissionCache` for cached role/permission resolution
- Update Postgres/OGC service registrations and JSON contexts
- Add store tests for WFS stored queries, workflow packages, and role caching

## Testing

## Gate Impact

## Docs or Contract Impact

## Release/Deploy Impact
- [x] Requires environment variable or secret changes

## Breaking Changes
None


mikemcdougall:
Closing: durable Redis-backed stores are net-new and were not part of the #1593 quality-sweep deliverables. Discarding for now; can revisit as a dedicated feature if needed.

Issue #2157, opened 2026-06-25
SIEM connector productization (Splunk/Sentinel/S3/syslog sinks) + audit retention/residency


mikemcdougall:
## Scope
Deferred slice of the audit/SIEM work in PR #2112. Productize SIEM connectors for Splunk, Microsoft Sentinel, S3, and syslog sinks, and add audit retention + data-residency controls.

## Acceptance Criteria
- Audit events can be exported to each sink (Splunk HEC, Microsoft Sentinel, S3, syslog) via configuration, with delivery retries/backoff and a dead-letter/failure path that does not drop the tamper-evident chain.
- Audit retention is configurable and enforced (events pruned after the window) while preserving tamper-evidence integrity for retained events.
- Data-residency controls constrain where audit data is stored/exported per configured region, with a test asserting an out-of-region sink is rejected.
- Integration tests cover at least one push sink (mocked) end-to-end plus retention pruning and residency enforcement.

## Effort
L.

## Dependencies
Built on PR #2112 (tamper-evident audit trail + SIEM export).

PR #2233, opened 2026-06-27
feat: add audit retention pruning background job (#509)

### LINKED ISSUE #509: SIEM export and operator access
Parent: #350

## Depends On

- #507 — Audit coverage matrix and middleware instrumentation

## Scope

Provide export, retention, and query surfaces so operators and SIEM systems can consume audit data.

- **Export endpoint**: JSON lines export via Admin API with configurable time-range filtering
- **Webhook/push integration**: configurable webhook sink to forward audit events to an external SIEM
- **Retention policy**: configurable retention period (e.g., 90 days) with automatic archival of older records
- **Admin query surface**: Admin API or UI read surface for paginated audit history queries filtered by actor, action, resource, or time range

## Out of Scope

- Audit event schema or storage (#504)
- Deciding which operations emit audit events (#507)
- Secure-connection-specific auditing (#354)
- Generic structured logging improvements

## Acceptance Criteria

- [ ] Audit events can be exported as JSON lines via an Admin API endpoint with time-range filtering
- [ ] A webhook sink can be configured to forward audit events to an external SIEM
- [ ] Retention policy is configurable (e.g., 90 days) with automatic archival of older records
- [ ] Admin API supports paginated audit history queries filtered by actor, action, resource, or time range

mikemcdougall:
# Pull Request

## Issue Link
Related to #509

## Summary
SIEM export and operator access (#509) was already largely implemented: the streaming JSON-Lines/CEF export endpoint (`GET /api/v1/admin/observability/audit/export`), the paginated/filterable history query surface, the tamper-evidence verifier, and the push sinks (Splunk HEC / Microsoft Sentinel / S3 / syslog) with shared retry/backoff, dead-lettering, and data-residency enforcement all landed under #2157.

The one acceptance criterion left unwired was **retention**: `AuditLog:Export:RetentionDays` existed and `AuditRetentionPolicy`/`IAuditRetentionPruner` were scaffolded, but no Postgres pruner and no background sweep were registered (the `IAuditRetentionPruner` doc and `AuditExportOptions` comment both said "deferred"). This PR implements automatic retention cleanup.

The pruner is integrity-safe by construction. The audit trail is append-only (migration 033's `audit_log_no_delete` `DO INSTEAD NOTHING` rule) and hash-chained (migration 069). The pruner lifts the delete guard inside a single transaction (restored on commit, and on rollback because DDL is transactional) and deletes only a **contiguous head prefix** of the chain. Because the integrity verifier treats the first surviving hashed row as the chain genesis (no `prev_hash` check on it), pruning the head never breaks verification — and by construction every row before the oldest non-expired row is itself expired, so no in-window row is removed.

## Changes Made
- Add `PostgresAuditLogRetentionPruner` (`IAuditRetentionPruner`) that prunes expired audit rows as a contiguous head prefix, preserving the tamper-evident chain.
- Add `AuditRetentionPruneService` background job that runs a bounded-retention sweep on a fixed interval, resolving the pruner in a scope and swallowing transient failures so a DB hiccup never crashes the host.
- Add `NullAuditRetentionPruner` no-op fallback for non-database / test hosts.
- Register the pruner + sweep in `AddHonuaAuditExport` when `RetentionDays > 0`, independent of whether SIEM push sinks are enabled.
- Update the now-accurate doc comments on `IAuditRetentionPruner` and `AuditExportOptions.RetentionDays`.
- Tests: Core no-op pruner unit tests; Postgres integration tests covering head-prefix pruning + chain re-verification, unbounded retain-forever, and append-only-guard restoration after a full clear.

## Testing
- [ ] Architecture tests pass

Ran:
- `dotnet build src/Honua.Server/Honua.Server.csproj -c Release /p:TreatWarningsAsErrors=true` → succeeded, 0 warnings / 0 errors (also builds Core + Postgres).
- `dotnet build tests/dotnet/Honua.Postgres.Tests -c Release /p:TreatWarningsAsErrors=true` → succeeded, 0 warnings.
- `dotnet test Honua.Core.Tests --filter NullAuditRetentionPruner` → 2 passed.
- `dotnet format --verify-no-changes` on all changed files → clean.

The new Postgres integration tests (`PostgresAuditLogRetentionPrunerTests`) compile and are Testcontainers-based; Docker is unavailable in this environment, so they run in CI rather than locally.

## Gate Impact

## Docs or Contract Impact

## Release/Deploy Impact

(No schema migration: retention reuses the existing `honua.audit_log` table and its append-only rule.)

## Breaking Changes
None

---

## Pre-PR Checklist

## Known gaps
- Postgres integration tests could not be executed locally (no Docker); they are validated in CI.
- Retention performs deletion (rotation), not copy-to-cold-storage "archival"; archival-before-delete can be layered on the existing S3 sink in a follow-up. The remaining #509 criteria (streaming export, webhook/SIEM forwarding sinks, paginated filtered query) were already implemented under #2157, so this PR uses `Related to #509`; close #509 once archival-vs-delete is confirmed acceptable.

### DIFF
--- a/src/Honua.Core/Features/AuditLog/Export/IAuditRetentionPruner.cs
+++ b/src/Honua.Core/Features/AuditLog/Export/IAuditRetentionPruner.cs
@@ -14,9 +14,11 @@ namespace Honua.Core.Features.AuditLog.Export;
 /// The audit trail is append-only and tamper-evident (a Postgres
 /// <c>DO INSTEAD NOTHING</c> rule blocks ordinary deletes). The production
 /// pruner is therefore a <em>privileged</em> maintenance path that bypasses the
-/// append-only rule under controlled conditions, and is deferred to the
-/// Postgres provider wiring. This Core seam exists so the retention policy and
-/// cutoff arithmetic can be unit-tested independently of a database.
+/// append-only rule under controlled conditions; the Postgres implementation
+/// (<c>PostgresAuditLogRetentionPruner</c>) lifts the delete guard inside a
+/// transaction and removes only a contiguous head prefix of the chain so
+/// integrity verification still passes. This Core seam exists so the retention
+/// policy and cutoff arithmetic can be unit-tested independently of a database.
 /// </para>
 /// </remarks>
 public interface IAuditRetentionPruner
--- a/src/Honua.Core/Features/AuditLog/NullAuditRetentionPruner.cs
+++ b/src/Honua.Core/Features/AuditLog/NullAuditRetentionPruner.cs
@@ -1,0 +1,29 @@
+// Copyright (c) Honua. All rights reserved.
+// Licensed under the Elastic License 2.0. See LICENSE in the project root.
+
+using Honua.Core.Features.AuditLog.Export;
+
+namespace Honua.Core.Features.AuditLog;
+
+/// <summary>
+/// No-op <see cref="IAuditRetentionPruner"/> used when no durable audit store is
+/// configured (tests and non-database hosts). Prunes nothing and reports zero.
+/// </summary>
+/// <remarks>
+/// This is intentionally <c>sealed</c> and stateless so DI can hand out a
+/// singleton without per-call allocation, mirroring <see cref="NullAuditLog"/>.
+/// A host with a real audit store registers a provider-backed pruner (for
+/// example the Postgres implementation) ahead of this fallback.
+/// </remarks>
+public sealed class NullAuditRetentionPruner : IAuditRetentionPruner
+{
+    /// <summary>A shared, allocation-free instance.</summary>
+    public static readonly NullAuditRetentionPruner Instance = new();
+
+    /// <inheritdoc />
+    public Task<int> PruneAsync(AuditRetentionPolicy policy, CancellationToken ct)
+    {
+        ArgumentNullException.ThrowIfNull(policy);
+        return Task.FromResult(0);
+    }
+}
--- a/src/Honua.Postgres/Features/AuditLog/PostgresAuditLogRetentionPruner.cs
+++ b/src/Honua.Postgres/Features/AuditLog/PostgresAuditLogRetentionPruner.cs
@@ -22,37 +22,57 @@ namespace Honua.Postgres.Features.AuditLog;
 /// deletes, and migration 069 builds a hash chain over the rows. Retention is the
 /// sanctioned, privileged maintenance path the migration explicitly allows
 /// ("rotate / archive via partition swap or explicit DROP RULE in a maintenance
-/// window"). This pruner performs that rotation safely:
+/// window"). This pruner performs that rotation safely and — critically —
+/// <em>without holding a long-lived lock</em>:
 /// </para>
 /// <list type="number">
-/// <item>It lifts the delete guard inside a single transaction and restores it in
-/// the same transaction, so the append-only rule is reinstated on commit — and a
-/// rollback (on any error) restores it too, because DDL is transactional in
-/// PostgreSQL.</item>
-/// <item>It deletes only a <em>contiguous prefix</em> of the chain (rows whose
-/// <c>audit_id</c> precedes the oldest still-retained row). Because the integrity
-/// verifier treats the first surviving hashed row as the chain genesis, pruning a
-/// head prefix never breaks verification — unlike a mid-chain delete. Only expired
-/// rows are removed: any row with <c>audit_id</c> below the first non-expired row's
-/// id is, by construction, itself expired.</item>
+/// <item>It establishes a single head-prefix safety boundary up front
+/// (<c>MIN(audit_id)</c> among still-retained rows). Everything below that
+/// boundary is, by construction, an expired contiguous prefix of the chain.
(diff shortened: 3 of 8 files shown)


mikemcdougall:
🤖 Automated batch review

Reviewed the retention pruner, background sweep, DI wiring, and tests against the surrounding audit-log code (migration `033_CreateAuditLog.sql`, `PostgresAuditLogIntegrityVerifier`, `AuditRetentionPolicy`, and the existing `AddHonuaAuditLog` registration).

**Verdict: LGTM with one operational should-fix.** The core design is correct: the `MIN(audit_id) WHERE timestamp >= cutoff` boundary guarantees only rows *before* the oldest retained row are deleted, so every deleted row is provably expired and the surviving rows remain a contiguous chain suffix. Because `PostgresAuditLogIntegrityVerifier` does not validate the first hashed row's `prev_hash` (genesis), a head-prefix prune cannot break verification — the integration test confirms this. The DI block correctly mirrors the established `AddHonuaAuditLog` pattern (resolve `IAdoNetDatabaseConnectionProvider`, else `Null*`), and placing the retention registration before the `if (!options.Enabled)` early return correctly makes retention independent of SIEM sinks.

### Should-fix
- **Lock duration on the first/large sweep** — `PostgresAuditLogRetentionPruner.PruneAsync` runs `ALTER TABLE … DISABLE RULE`, the full `DELETE`, and `ENABLE RULE` in a single transaction. `ALTER TABLE … DISABLE/ENABLE RULE` takes an `ACCESS EXCLUSIVE` lock, held from the DISABLE statement until COMMIT. That blocks **all** concurrent `honua.audit_log` inserts (and reads) for the entire delete. `RecordAsync` is awaited inline in request handlers (e.g. `OperationGateway`, the admin endpoints), so on a large expired backlog the first sweep can stall audit-writing requests for as long as the bulk delete runs. Consider deleting in bounded chunks across several short transactions (disable → delete N → enable → commit, loop), or at minimum a `statement_timeout`/`LIMIT`-style cap so one sweep can't hold the table exclusively for an unbounded period. Correctness is fine; this is purely about lock-hold time.

### Nits
- **Provider coupling** (`AuditExportServiceCollectionExtensions.cs`): the pruner is constructed whenever any `IAdoNetDatabaseConnectionProvider` is present, but `DuckDB`/`MySql` also register that interface, and the pruner emits Postgres-specific DDL (`DISABLE RULE`). This mirrors the pre-existing `AddHonuaAuditLog` → `new PostgresAuditLog(...)` behavior (so it's consistent, and a failed sweep is swallowed by the background service), but it confirms audit storage is implicitly Postgres-only. A non-Postgres host with `RetentionDays > 0` will log a swallowed warning every 6h. Worth a guard on the active provider if non-Postgres audit hosts are ever supported.
- **Out-of-order timestamps**: head-prefix-only pruning leaves an expired row with `audit_id > keep_from` (i.e. a late-arriving older `Timestamp`) un-pruned indefinitely. Fine in practice since `audit_id` tracks insert order and `Timestamp` is ~insert time — and it's documented — but it means retention is best-effort, not a hard purge guarantee (relevant if used for a GDPR/erasure claim).
- **Test schema duplication** (`PostgresAuditLogRetentionPrunerTests.EnsureAuditLogTableAsync`): the table + rules are hand-rolled rather than applied from migrations `033`/`069`. It currently matches, but it will silently drift if the real schema changes; running the actual migration would be more robust.

mikemcdougall:
🤖 Automated fix: Addressed the review finding on the audit retention pruner.

`PostgresAuditLogRetentionPruner` no longer does `DISABLE RULE` + full `DELETE` + `ENABLE RULE` inside a single transaction (which held an `ACCESS EXCLUSIVE` lock on `honua.audit_log` until COMMIT and blocked all inline audit inserts during a large first sweep). It now:

- Establishes the head-prefix safety boundary once up front (`MIN(audit_id)` among still-retained rows), so only provably-expired contiguous head rows are removed and hash-chain verification still holds.
- Deletes expired rows in bounded chunks (`DELETE ... WHERE ctid IN (SELECT ctid ... ORDER BY audit_id LIMIT batchSize)`), each in its own SHORT transaction, looping until none remain.
- Lifts/restores the append-only guard per batch, so the `ACCESS EXCLUSIVE` lock is held only for milliseconds per batch and released between batches — inline audit writes interleave with pruning instead of stalling.
- Makes batch size configurable: `AuditLog:Export:RetentionPruneBatchSize` (default 5000).

Tests: added `PruneAsync_SmallBatchSize_DeletesExpiredPrefixAcrossMultipleBatches_AndRestoresGuard` (batchSize=2 over 5 expired rows) asserting multi-batch deletion, retained recent rows, chain still verifies, and the guard is restored. `Honua.Server` and `Honua.Postgres.Tests` build clean in Release (0 warnings). The retention integration tests require Docker/Testcontainers, which was unavailable in this environment, so they were not executed here.

mikemcdougall:
Superseded by the consolidated batch bundle #2259 — bundling all batch tickets into one PR for a single CI run.

mikemcdougall:
Folded into the consolidated bundle PR #2259 (one PR / one CI). Closing as superseded; the commits are merged into `codex/batch-bundle-2026-06-27`.

Issue #2229, opened 2026-06-27
Implement GML file import (reader + format detection)


mikemcdougall:
﻿## Context
The original GML issue (#1595) was closed by [PR #1605](https://github.com/honua-io/honua-server/pull/1605), which **aligned the advertised import formats by removing GML** rather than implementing it. As a result, genuine GML import support is still absent from `trunk` (`FileFormatDetectionService` does not detect GML).

## Ask
Implement real GML file import:
- GML format detection in `FileFormatDetectionService`
- A `GmlFormatReader` wired through the streaming import pipeline (CRS detection, preview, streaming)
- Re-advertise GML in the supported import formats once functional
- Unit tests for detection and reading

## Notes
Implementation proposed in PR #2227 (rebased onto current trunk).

PR #2295, opened 2026-06-29
test: stop asserting GML is an unsupported import format

mikemcdougall:
## Summary

GML import support (`SupportedFileFormat.Gml`, the `.gml` extension, and content sniffing for `opengis.net/gml`) landed via #2229, but two tests were never updated and still asserted that `.gml` is an **unsupported / unadvertised** format. Those stale assertions fail the **Foundation** and **FileImport** test shards, which is the root cause of CI being red on `trunk` (and therefore on every PR).

Positive GML coverage already exists in `GmlFormatDetectionTests`; this change only removes the now-incorrect negative assertions, keeping the still-unsupported TWKB checks.

## Changes Made

- `FileGdbFormatDetectionTests.DetectFormat_UnsupportedAdvertisedFormats_AreRejected`: drop the `.gml` rejection assertions; keep `.twkb`.
- `ImportEndpointTests.GetSupportedFormats_DoesNotAdvertiseUnsupportedFormats`: drop the `.gml` not-advertised assertions; keep `.twkb`.

## Breaking Changes

None. Test-only change.

## Gate Impact

- [x] Tests only — fixes existing failing tests; no product behavior change.

## Testing

- [x] Verified GML is genuinely supported on trunk (enum value, `.gml` extension map, content sniffing) and that TWKB remains unsupported; positive GML detection is covered by `GmlFormatDetectionTests`.

### DIFF
--- a/tests/dotnet/Honua.Core.Tests/Features/Import/FileGdbFormatDetectionTests.cs
+++ b/tests/dotnet/Honua.Core.Tests/Features/Import/FileGdbFormatDetectionTests.cs
@@ -28,9 +28,7 @@ public sealed class FileGdbFormatDetectionTests
     {
         var service = new FileFormatDetectionService(NullLogger<FileFormatDetectionService>.Instance);
 
-        service.DetectFormat("sample.gml").Should().BeNull();
         service.DetectFormat("sample.twkb").Should().BeNull();
-        service.GetSupportedExtensions().Should().NotContain(".gml");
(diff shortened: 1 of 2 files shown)


mikemcdougall:
Folded into the consolidated trunk-CI-greening PR #2298 (one PR / one CI run). Closing as superseded; the commits are merged into `fix/green-trunk-ci`.

PR #2296, opened 2026-06-29
fix: GP jobs trailing-slash guard, import-staging name-length migration (071), MCP publish_service conformance

mikemcdougall:
Rescued a coherent bundle of small fixes that had no branch identity. None of these had landed on trunk (verified against origin/trunk file-by-file; migration 071 and the publish_service schema are new).

## Changes

- **GeoServices GPServer trailing-slash jobId guard** (`GPServerEndpoints.cs`): A request to `.../jobs/` (trailing slash, empty `{jobId}`) is collapsed by ASP.NET routing onto the jobs-listing route. An Esri client addressing a specific job with an empty/missing id now gets a `Missing jobId.` bad-request error instead of the full job listing.

- **Migration 071 — import staging table name length** (`071_FixImportStagingTableNameLength.sql`, new): Migration 070 derived the staging sibling name as `<table>__staging` and hard-failed (P0001) when that exceeded PostgreSQL's 63-character identifier limit, so a valid target name >= 55 chars could never be imported in default replace mode. Adds a shared, deterministic `honua.import_staging_table_name` helper bounded to <= 48 chars (readable form for short names, truncated prefix + md5 suffix for long ones) used by both `create_import_staging_table` and `swap_import_table` so the computed names always agree.

- **MCP geospatial-mcp conformance — publish_service** (`index.json`, `tools/publish_service.schema.json` new, `McpGeospatialMcpSchemaConformanceTests.cs`): Adds the `publish_service` tool schema and registers `honua_publish_service` -> `publish_service` in the taxonomy alignment map.

- **TestKit DROP DATABASE race** (`PostgresFixture.cs`): Flush process-wide Npgsql pools (`ClearAllPools`) and use `DROP DATABASE ... WITH (FORCE)` to close the teardown race where reconnecting pooled backends caused "database is being accessed by other users".

## Test results

- `Honua.Server` (and its `Honua.Protocols.GeoServices` dependency) build: succeeded, 0 warnings/errors.
- MCP conformance tests (`McpTaxonomyAlignmentTests`, which exercise the new schema/index): 35 passed, 0 failed.
- GP endpoint integration tests and PostgresFixture-backed tests could not be executed in this environment (no Docker for Testcontainers and no external `HONUA_TEST_DB_URL`); they should run in CI.


mikemcdougall:
Folded into the consolidated trunk-CI-greening PR #2298 (one PR / one CI run). Closing as superseded; the commits are merged into `fix/green-trunk-ci`.

PR #2358, opened 2026-07-02
fix: FileImport shard SASL no-password + stale staging-table seed

### LINKED ISSUE #1628: OGC MVT serves stale publish-time materialization while FeatureServer reads live tables — re-imports silently stale
Found seeding the hosted demo: FeatureServer queries read the live typed tables, but OGC API Tiles MVT reads the publish-time `public.features` materialization. After re-importing/updating a layer, FeatureServer shows new data while MVT keeps serving the old — silently. Consequence: any update workflow (or the demo's nightly reseed) leaves tiles stale until an explicit re-publish/flatten. Request: MVT should read the same source as FeatureServer, or re-import must invalidate/rebuild the materialization automatically. Data-consistency class bug.

mikemcdougall:
# Pull Request

## Issue Link
Fixes  #1628 follow-up (FileImport shard SASL)

## Summary
Fixes the persistent `Server Tests (FileImport)` integration-shard failure that was re-escalating the honua-server release merge-train batch. The shard exhibited two distinct, reproducible problems that this PR resolves at the harness/config level without weakening authentication:

1. **`Npgsql: No password has been provided but the backend requires one (SASL/SCRAM-SHA-256)`** — the post-import canonical-snapshot refresh (`ImportEndpoints.TryRefreshPublishedSnapshotAsync` -> `PostgreSqlLayerPublishingService.RefreshMaterializedFeaturesForSourceTableAsync`) opens a *fresh* `NpgsqlConnection` from `IDatabaseConnectionProvider.GetConnectionString()`. That method returns `NpgsqlDataSource.ConnectionString`, which Npgsql redacts the password from by default (`Persist Security Info=false`), so the reconnect failed SCRAM auth. This is a **real product bug**, not test-only: in production the same path silently swallows the failure as a warning, leaving published MVT tiles stale after an import (defeating honua-server#1628). Fixed by setting `PersistSecurityInfo=true` at the single datasource-build site so the exposed string stays usable — the backend still requires the password, so auth is not weakened.

2. **`P0001: Staging table name exceeds PostgreSQL identifier limit of 63 characters`** (the assertions that actually turned the shard red — `ImportFile_CustomTargetSchema_CanDiscoverAdHocImportSchema` and `ImportFile_DefaultOperationalSchema_CanDiscoverPublishAndQueryFeatureServer`). The migration-skipping integration fixture seeds its `honua.*` import functions from `tests/seed/server.yaml`, whose copy of `honua.create_import_staging_table` had drifted: it still RAISEd on long `<table>__staging` names — the exact behavior migration `071_FixImportStagingTableNameLength.sql` replaced with a length-safe md5 fallback. The seed was never updated when 071 landed. Fixed by mirroring migration 071 in the seed.

## Changes Made
- `PostgresDataSourceFactory.Configure`: set `PersistSecurityInfo=true` so `NpgsqlDataSource.ConnectionString` (surfaced via `IDatabaseConnectionProvider.GetConnectionString()`) retains the password for reconnect-from-string paths (snapshot refresh, layer-publish materialize).
- `WebAppFixturePostgresWiringMixin.OverrideNonMultiplexingDataSource`: mirror the same `PersistSecurityInfo=true` on the test-overridden datasource so the integration host reproduces production behavior.
- `tests/seed/server.yaml`: add the `honua.import_staging_table_name` helper and switch `honua.create_import_staging_table` / `honua.swap_import_table` to the length-safe md5 fallback, mirroring migration 071 (removes the P0001 RAISE).
- Add `PostgresDataSourceFactoryTests.Create_ExposedConnectionString_RetainsPasswordForReconnectPaths` regression test.

## Testing
- [ ] Architecture tests pass

Verified locally (Release, Testcontainers PostGIS):
- `PostgresDataSourceFactoryTests`: 31/31 passed (incl. the new password-retention test).
- The two previously-failing FileImport tests + the snapshot-refresh (`GeoParquetImportTests.Upload_WithNestedColumn_...`) test: 3/3 passed, with zero `No password`/`SASL`/`P0001`/`SnapshotRefreshFailed` signatures in the run log.

## Gate Impact

## Docs or Contract Impact

## Release/Deploy Impact

## Breaking Changes
None

---

## Pre-PR Checklist

### DIFF
--- a/src/Honua.Postgres.Shared/Features/Infrastructure/PostgresDataSourceFactory.cs
+++ b/src/Honua.Postgres.Shared/Features/Infrastructure/PostgresDataSourceFactory.cs
@@ -64,6 +64,18 @@ internal static class PostgresDataSourceFactory
 
         connectionStringBuilder.Pooling = true;
 
+        // Retain the password in the string exposed by NpgsqlDataSource.ConnectionString
+        // (surfaced through IDatabaseConnectionProvider.GetConnectionString()). The post-import
+        // canonical-snapshot refresh and layer-publish materialize paths take that string and
+        // open a *fresh* NpgsqlConnection from it. Npgsql strips the password by default
+        // (Persist Security Info=false), so those fresh connections failed SCRAM auth with
+        // "No password has been provided but the backend requires one (in SASL/SCRAM-SHA-256)"
+        // — swallowed as a warning in TryRefreshPublishedSnapshotAsync, leaving published MVT
+        // tiles silently stale after an import (honua-server#1628). This does NOT weaken auth:
+        // the backend still requires the password; it only stops Npgsql from redacting it from
+        // the string handed back to code that must reconnect.
+        connectionStringBuilder.PersistSecurityInfo = true;
+
         if (limits.MaxConnectionPoolSize > 0)
         {
             connectionStringBuilder.MaxPoolSize = limits.MaxConnectionPoolSize;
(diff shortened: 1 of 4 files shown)


mikemcdougall:

Issue #2363, opened 2026-07-02
bug(formats): .wkt import cannot parse EWKT (SRID= prefix) or multi-line geometries — yields 0 features


mikemcdougall:
## Problem Summary

The `.wkt` file import reader (`WktFormatReader`) reads exactly one geometry per line via NTS `WKTReader.Read(line.Trim())` and silently skips any line that fails to parse. It therefore cannot import (a) EWKT values carrying an `SRID=…;` prefix (the PostGIS `ST_AsEWKT` form) or (b) a single geometry pretty-printed across multiple lines. Invalid/unrecognized lines are swallowed by a bare `catch {}`, so an all-EWKT file yields **0 features** and the import fails with "No features found in file," while a mixed file silently drops the multi-line geometries and reports success.

## Why This Matters Now

`.wkt` is a shipped import format. The CRS-detection path for WKT is written specifically around the EWKT `SRID=…;` prefix (`StreamingFileImportService` `_wktSridRegex = "SRID\\s*=\\s*(\\d+)\\s*;"`, `CrsDetection.DetectWktSridAsync`), so the pipeline anticipates EWKT input the reader then cannot parse — an internal contradiction, not a design choice. Users exporting WKT from PostGIS (`ST_AsEWKT`) get a confusing "no features" failure on a file that plainly contains geometries; mixed one-per-line/multi-line files silently lose rows. Same NTS-`WKTReader`-can't-parse-EWKT root cause as the CSV silent-loss issue (#2361).

## Steps to Reproduce

1. Create `data.wkt` with one EWKT geometry per line: `SRID=4326;POINT (30 10)`.
2. Import it. SRID 4326 is detected, but every `WKTReader.Read` throws → all lines skipped → import fails: "No features found in file."
3. Alternatively, put a single geometry across multiple lines (pretty-printed `POLYGON ((...\n...))`) → dropped.

## Expected Behavior

EWKT (`SRID=…;`) values import with the detected SRID honored (strip the prefix as `CrsDetectionService.cs:164-166` / `ODataFilterParser.cs:641` already do), and multi-line geometries are accepted. A file that contains valid geometries must not report "No features found."

## Actual Behavior

[...]

PR #2367, opened 2026-07-02
fix(formats): parse EWKT/EWKB in CSV and WKT importers, warn on unparseable CSV geometry

### LINKED ISSUE #2361: bug(formats): CSV import silently loses all geometry (EWKT/EWKB-hex) and drops shape/geom/geometry columns — reports success
## Problem Summary

CSV import can silently lose all geometry **and** silently drop a source column while reporting the import as `Completed` with zero warnings. Two distinct defects in `CsvFormatReader` combine:

1. A CSV column whose (normalized) name is `wkt`, `geomwkt`, `geometrywkt`, `thegeom`, `geom`, `geometry`, or `shape` is claimed as the geometry column and **unconditionally excluded from the attribute table** — even when its values are not parseable WKT.
2. When the claimed geometry value fails to parse as WKT, the failure is swallowed by a bare `catch {}` and the row is imported with **NULL geometry**; the raw source value is not retained anywhere. Only GeoParquet skips+warns on null geometry — CSV (and other vector formats) do not.

Net effect: a valid, shipping-format import completes with HTTP 200 / status `Completed` while every feature has NULL geometry and the source column has vanished. This is exactly the "reports success while losing data" silent-failure class (cf. #2352/#2353/#2354).

## Why This Matters Now

The most common way to get tabular spatial data out of PostGIS/other databases into a CSV is a `\copy`/`COPY ... TO` or `ST_AsEWKT`/`ST_AsText` export, which produces a geometry column typically named `geom`/`geometry`/`wkt` containing **EWKB hex** (e.g. `0101000020E6100000...`) or **EWKT** (`SRID=4326;POINT(...)`). NTS `WKTReader` cannot parse either form, so importing such a CSV yields a layer where **every feature has NULL geometry**, silently — the user sees "import completed" and only discovers the total spatial-data loss later (blank map / empty query results). Separately, any benign non-spatial CSV with an attribute column literally named `shape`, `geom`, or `geometry` has that column silently deleted from every row. Silent data loss on a shipped, advertised import format is a release-gate defect.

## Steps to Reproduce

A) EWKT/hex geometry (PostGIS export):
1. Export a table to CSV so the geometry column (`geom`) holds EWKB hex or `SRID=4326;POINT(30 10)`.
2. `POST /api/.../import` the CSV.
3. Observe status `Completed`, `FeaturesProcessed = N`, no warnings — but every row has NULL geometry and the `geom` column is absent from attributes.

B) Column-name collision:
1. Take a non-spatial CSV with a column named `shape` (values like `round`, `square`).
2. Import it. Observe that the `shape` column is silently dropped from every imported row (recognized as a geometry column, excluded from attributes, and — being unparseable — produces null geometry).

## Expected Behavior

- A geometry value that fails to parse should either fail the row (counted as failed) or be surfaced as a warning — never silently produce NULL geometry with a `Completed`/no-warning result.
- The raw source value should not be silently discarded when it is not usable as geometry.
- Rows imported with NULL geometry on a spatial import should be surfaced (skipped+counted or warned), consistent with the GeoParquet path (`#423`).
- EWKT (`SRID=...;`) values should be honored (strip the SRID prefix as `CrsDetectionService.cs:164-166` and `ODataFilterParser.cs:641` already do) or, if hex EWKB, decoded — not silently nulled.

## Actual Behavior

`src/Honua.Geometry/Features/FileImport/Services/CsvFormatReader.cs`:
- `:37-47` `_wktColumnNames` includes `shape`, `geom`, `geometry`, etc.
- `:227-230` first column matching a WKT name becomes `WktIndex`.
- `:153-154` + `:248-249` `IsGeometrySourceColumn` → `continue;` excludes that column from attributes unconditionally.
- `:182-195` `ParseGeometry` wraps `wktReader.Read(...)` in a bare `catch {}` → returns null on any parse failure.
- `:167-173` a feature with null geometry but any attributes is still returned (imported).

Import wiring / success reporting:
- Dispatch: `src/Honua.Postgres/Features/FileImport/StreamingFileImportService.Streaming.cs:108`.
- Null geometry → `CreateWkb` returns null and the row is inserted with NULL geometry (`StreamingFileImportService.Batch.cs:361-364`).
- Null-geometry skip+warn is applied **only** for GeoParquet (`StreamingFileImportService.Streaming.cs:121, 216`); CSV gets neither.
- Completion is reported as `ImportStatus.Completed` unconditionally (`StreamingFileImportService.Streaming.cs:229-249`).

Root cause shared with the WKT-reader issue: NTS `WKTReader` cannot parse EWKT (`SRID=…;`) / EWKB-hex, and the file-import readers never strip/decode it — even though the CSV/WKT CRS detector (`_wktSridRegex`, `CrsDetection.DetectWktSridAsync`) is built around exactly that `SRID=` prefix.

## Acceptance Criteria

- [ ] Importing a CSV whose geometry column contains EWKT or EWKB-hex either imports the geometry correctly or fails/warns visibly — never a silent `Completed` with all-NULL geometry.
- [ ] A column recognized as geometry that fails to parse does not silently disappear (either surfaced as an attribute or reported in warnings).
- [ ] Null-geometry rows on CSV import are surfaced (skipped+counted or warned), consistent with GeoParquet.
- [ ] Regression tests cover EWKT and EWKB-hex CSV geometry columns and the `shape`/`geom`/`geometry` column-name-collision case.

## Affected Repo(s)

honua-server

## Gate Tier Impact

Nightly gates (conformance, performance, security)

## Release/Deploy Impact

None — standard merge-and-release

## Environment

- Honua version/commit: trunk @ 264a7d381
- Deployment: any (PostgreSQL/PostGIS backend)
- Discovered via read-only code audit against origin/trunk.

## Non-Goals

- Full GDAL-grade format sniffing; this is about not silently losing data on the existing CSV reader path.

mikemcdougall:
# Pull Request

## Issue Link
Closes #2361 #2363

## Summary
Standard PostGIS CSV/`.wkt` exports encode geometry as EWKT (`SRID=<n>;WKT`) or WKB/EWKB hex, but the CSV and WKT format readers only understood plain OGC WKT. As a result:

- **#2361 (S1):** a CSV column named `wkt`/`geom`/`geometry`/`shape` was claimed as the geometry source, but if its value was EWKT or EWKB-hex a bare `catch{}` produced a NULL geometry, discarded the raw value, and imported the row as `Completed` with zero warnings — silent, total geometry loss on the standard export form.
- **#2363 (S2, shared root cause):** `.wkt` import could not parse EWKT (despite the WKT CRS-detector being built around the `SRID=` prefix) or multi-line geometries, yielding 0 features.

This PR fixes both via a shared geometry-text parser and makes CSV geometry-parse failures non-silent.

## Changes Made
- Add `GeometryValueParser` (Honua.Geometry) — a shared helper that parses plain WKT, EWKT (stripping and honouring the `SRID=` prefix), and WKB/EWKB hex, so the CSV and WKT readers cannot diverge on accepted encodings.
- `WktFormatReader`: accumulate parenthesis-balanced records so multi-line geometries are parsed as one unit, and parse EWKT/EWKB via the shared helper (#2363).
- `CsvFormatReader`: parse the geometry column via the shared helper; when a mapped geometry column holds a value that cannot be parsed by any encoding, preserve the raw value as an attribute (never discarded) and record it on a `CsvGeometryDiagnostics` sink instead of silently dropping to NULL (#2361).
- `StreamingFileImportService`: surface an import warning when a CSV had unparseable-geometry rows, mirroring the existing GeoParquet null-geometry skip/warn path.
- Add regression tests: PostGIS-export CSVs (EWKT and EWKB-hex) import with geometry + featureCount>0; an unparseable CSV geometry value produces a diagnostic (not silent null-success) with the raw value preserved; `.wkt` files with `SRID=4326;POINT(...)`, WKB hex, and multi-line WKT import features.

## Testing
- [x] Architecture tests pass

## Gate Impact
- [ ] None — no gate impact

## Docs or Contract Impact

## Release/Deploy Impact

## Breaking Changes
None

---

## Pre-PR Checklist

### DIFF
--- a/src/Honua.Geometry/Features/FileImport/Services/CsvFormatReader.cs
+++ b/src/Honua.Geometry/Features/FileImport/Services/CsvFormatReader.cs
@@ -70,14 +70,29 @@ internal static class CsvFormatReader
     internal static IAsyncEnumerable<IFeature> ReadStreamingAsync(
         Stream stream,
         CancellationToken cancellationToken)
-        => ReadStreamingAsync(stream, delimiterOverride: null, cancellationToken);
+        => ReadStreamingAsync(stream, delimiterOverride: null, diagnostics: null, cancellationToken);
 
     /// <summary>
     /// Stream features from a CSV file with an optional explicit delimiter override.
     /// </summary>
+    internal static IAsyncEnumerable<IFeature> ReadStreamingAsync(
+        Stream stream,
+        char? delimiterOverride,
+        CancellationToken cancellationToken)
+        => ReadStreamingAsync(stream, delimiterOverride, diagnostics: null, cancellationToken);
+
+    /// <summary>
+    /// Stream features from a CSV file with an optional explicit delimiter override and an
+    /// optional <see cref="CsvGeometryDiagnostics"/> sink. When a row has a mapped geometry
+    /// column whose value is present but cannot be parsed as WKT/EWKT/WKB, the raw value is
+    /// preserved as an attribute (so the data is never discarded) and the occurrence is
+    /// recorded on <paramref name="diagnostics"/> so the caller can surface a warning rather
+    /// than silently importing a null geometry.
+    /// </summary>
     internal static async IAsyncEnumerable<IFeature> ReadStreamingAsync(
         Stream stream,
         char? delimiterOverride,
+        CsvGeometryDiagnostics? diagnostics,
         [EnumeratorCancellation] CancellationToken cancellationToken)
     {
         using var reader = new StreamReader(stream, Encoding.UTF8, detectEncodingFromByteOrderMarks: true, leaveOpen: true);
@@ -86,6 +101,7 @@ internal static class CsvFormatReader
         CsvColumnMapping mapping = default;
         var geometryFactory = new GeometryFactory(new PrecisionModel(), 4326);
         var wktReader = new WKTReader();
+        var wkbReader = GeometryValueParser.CreateWkbReader();
         var sampleRecords = await ReadSampleRecordsAsync(reader, cancellationToken);
         var delimiter = delimiterOverride ?? DetectDelimiter(sampleRecords);
 
@@ -104,7 +120,7 @@ internal static class CsvFormatReader
                 return null;
             }
 
-            return BuildFeature(headers, fields, mapping, geometryFactory, wktReader);
+            return BuildFeature(headers, fields, mapping, geometryFactory, wktReader, wkbReader, diagnostics);
         }
 
         foreach (var sampleRecord in sampleRecords)
@@ -144,7 +160,9 @@ internal static class CsvFormatReader
         IReadOnlyList<string> fields,
         CsvColumnMapping mapping,
         GeometryFactory geometryFactory,
-        WKTReader wktReader)
+        WKTReader wktReader,
+        WKBReader wkbReader,
+        CsvGeometryDiagnostics? diagnostics)
     {
         var attributes = new AttributesTable();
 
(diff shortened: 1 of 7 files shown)


mikemcdougall:

PR #2394, opened 2026-07-03
fix: harden generateToken against credential exposure in query strings (PA-059)

mikemcdougall:
## Summary

- PA-059 S1 security fix: passwords sent via URL query string to GET /sharing/rest/generateToken are recorded verbatim in web server, reverse-proxy, and load-balancer access logs
- Real ArcGIS/Esri clients use GET with credentials in the query string, so a blanket 400 rejection would break compatibility. This fix preserves the GET path while hardening it.

## Changes

SharingRestEndpoints.cs: When RequireHttps=true (the default) and a GET request carries credentials in the URL over plain HTTP, the 403 now says "Credentials in URL query strings are not safe over plaintext HTTP. Use POST with a form-encoded body, or enable HTTPS." The GET endpoint WithDescription also now explicitly recommends POST.

PortalTokenLog.cs: Added EventId 7007. Every generateToken request that delivers credentials via the query string emits a Warning-level log event. The password value is never written; only the channel is noted.

Program.cs: UseSerilogRequestLogging now pins the MessageTemplate to the path-only format. This ensures query strings — including password and token values on the generateToken endpoint — can never reach log sinks even if a future enricher or template change would otherwise expose them.

## How HTTPS is detected

IsHttpsRequest calls context.Request.IsHttps. ASP.NET Core resolves this correctly after the trusted ForwardedHeaders middleware applies X-Forwarded-Proto.

## Redaction mechanism

Serilog default {RequestPath} already excludes query strings; pinning the template makes this an explicit, auditable guarantee. Application-level logging uses LogValueRedactor.Hash for sensitive values. EventId 7007 records only the credential channel, never the password value.

## Test plan

- [x] GenerateToken_CredentialsInQueryStringOverPlainHttp_Returns403WithHelpfulMessage
- [x] GenerateToken_CredentialsInQueryStringWithInsecureMode_Succeeds
- [x] All 17 pre-existing SharingRestTokenTests pass
- [x] dotnet build Honua.Protocols.GeoServices: 0 warnings, 0 errors
- [x] dotnet build Honua.Server: 0 warnings, 0 errors

Fixes PA-059.

### DIFF
--- a/src/Honua.Hosting/Features/Authentication/PortalTokenLog.cs
+++ b/src/Honua.Hosting/Features/Authentication/PortalTokenLog.cs
@@ -39,4 +39,16 @@ internal static partial class PortalTokenLog
     [LoggerMessage(EventId = 7006, Level = LogLevel.Warning,
         Message = "Portal token store could not remove distributed cache entry {KeyHash}.")]
     public static partial void DistributedCacheRemoveFailed(ILogger logger, string keyHash, Exception exception);
+
+    /// <summary>
+    /// Emitted when a <c>generateToken</c> request carries credentials via the URL query
+    /// string rather than a POST form body. The password value is never included in the
+    /// log line; this entry is an audit signal for operators monitoring for insecure
+    /// credential transmission patterns.
+    /// </summary>
+    [LoggerMessage(EventId = 7007, Level = LogLevel.Warning,
+        Message = "Portal token request received credentials via URL query string; " +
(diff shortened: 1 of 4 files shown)


mikemcdougall:

mikemcdougall:
Superseded by #2405 (consolidated into one PR for a single CI run per the audit consolidation). All commits from this branch are merged into the consolidation branch; closing to keep one PR/one CI.

PR #2423, opened 2026-07-04
Feature bug-hunt Round 1: 32 verified bugs across non-experimental features (S0 + 11 S1 + 20 S2)

mikemcdougall:
Round 1 of the feature-area bug-bounty of non-experimental honua-server features. 44 raw findings → 32 survived double adversarial verification (each confirmed by 2 independent refuters) + my line-by-line validation of the S0/S1s. All fixed with regression tests; consolidated for one CI run. Experimental features (SensorThings, Federated Query, Redshift/Snowflake/Databricks, mTLS, alerts/geofence, realtime, sync, temporal.*) were out of scope.

## S0
- **JWT replay-protection infinite recursion** (`AtomicTokenReplayProtection`): the Redis atomic-path fallback called back into the dispatcher which re-dispatched to Redis → StackOverflow on auth. Fixed: fallback now goes to a terminal non-atomic path; SOE regression test.

## S1 (11)
- **applyEdits/deleteFeatures authz bypass** (×2): a principal with Update-but-not-Delete could delete/insert features (only Update was checked). Now gates deletes on Delete and adds on Insert, per-payload + pre-body.
- **WFS Transaction cluster** (×4): partial-commit despite `rollbackOnFailure=true` (now fail-safe rejects the un-atomic multi-layer case), paging links dropping all filters (page 2 wrong set), zero-match → 400 instead of no-op, 3D `gml:posList` flattening coords.
- **STAC /queryables authz bypass**: leaked the field schema of access-protected collections without a credential — now enforces RequireResourceAccess.
- **Bidirectional replica sync** dropping concurrent server edits.
- **GP numeric filter** (`CompareNumeric`): `int.MinValue < 0` made `lt`/`lte` pass every non-numeric row — now returns `int?`/null so non-numeric fails all comparisons.
- **GP path-traversal**: arbitrary server file read via a user-injected stream-reference path — now containment-checked against the output root.
- **Zarr unbounded per-chunk buffer** (OOM) — now capped.

## S2 (20)
Across GeoServices (idempotency key collision, queryAttachments filter bypass, null-geometry create), OGC API (GML numberReturned, paging key encoding, PATCH geometryChanged), providers (SqlServer cross-SRID, DuckDB unknown-column), raster/tiles (PMTiles lon/lat clamp, ESRI tile-size logical shift, MapAlgebra exponent), STAC (open-interval datetime, I3S Oid32 truncation), OData (Layers skiptoken fingerprint), auth (session-store Redis-failover 401s), and a TOCTOU upload race. Each fixed per its finding; regression tests where practical.

Also fixed 5 pre-existing GDAL raster test failures (stale `gdalinfo` invocation assertion). Register: ~/honua-io/_bughunt/register/round1-survivors.jsonl.

### DIFF
--- a/src/Honua.Core/Features/Edit/EditProcessor.cs
+++ b/src/Honua.Core/Features/Edit/EditProcessor.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Honua. All rights reserved.
+﻿// Copyright (c) Honua. All rights reserved.
 // Licensed under the Elastic License 2.0. See LICENSE in the project root.
 
 using System.Collections.Immutable;
@@ -450,6 +450,13 @@ public sealed class EditProcessor : IEditProcessor
             if (feature.Attributes?.IsEmpty != false && RequiresAttributes(resource))
             {
                 return EditValidationResult.Failure("Attributes required for create operation");
+            }
+
+            // BH-014: Validate geometry exists if the layer is spatial so the error surfaces
+            // as a clean validation failure rather than a DB NOT NULL constraint violation.
+            if (feature.Geometry is null && RequiresGeometry(resource))
+            {
+                return EditValidationResult.Failure("Geometry is required for create operation on a spatial layer");
             }
         }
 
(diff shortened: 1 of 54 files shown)


mikemcdougall:

PR #2467, opened 2026-07-05
feat(ops): close local-disk statelessness leaks - checkpoints, import staging, scene tilesets

mikemcdougall:
Closes #2459. Related to #2457 (ADR-0060 WS0).

## Summary

Closes the local-disk statelessness leaks so compute holds nothing irreplaceable — the prerequisite for disposable/rolling deploys (ADR-0060). Three leaks, three fixes:

**Migration-run checkpoints → Postgres.** `IMigrationRunCheckpointStore` had no production DI wiring at all; the only durable implementation wrote resumable cross-request checkpoints to the node's local disk. A new `PostgresMigrationRunCheckpointStore` (jsonb upsert keyed by `run_id`, sanitizer applied on write, AOT-safe source-gen JSON) becomes the durable path with migration `073`; the in-memory store is the registered fallback (as a process-wide singleton — a per-scope instance would silently drop checkpoints between requests) and the filesystem store is documented dev/test-only.

**Import upload staging → within-request scratch contract.** The staged upload temp file was traced end-to-end: the synchronous path reads and deletes it inside the same request (legitimate scratch), and the background path already externalizes to `ICloudFileStorage`. The one latent leak — queuing a background import that silently kept depending on node-local temp when no cloud handoff was produced — is closed: background imports now require the object-store handoff (503 otherwise) and the local temp is always deleted. The dead `StreamingFileUploadService.QueueFileUploadAsync` path (zero callers) was removed while preserving the `IUploadQueueMetricsProvider` surface that health checks and monitoring consume.

**Scene tilesets → object store with read-through hydration.** Hosted 3D Tiles scenes were published to and served off one node's disk (`asset_root`), so a scene published on node A was unservable from node B. Publish now uploads the promoted tileset tree to `ICloudFileStorage` under deterministic `scenes/{sceneId}/…` keys (manifest uploaded last, so a readable manifest guarantees a complete tree) and records an `asset_storage_prefix` (migration `074`); serving hydrates through a read-through local materialization cache (`SceneAssetHydrator`: per-scene single-flight download, atomic install, content-version marker so republish re-hydrates). The hydration seam sits in the registry's `FindAsync` projection, so all six serving sites (HTTP, I3S, OpenUSD, gRPC) and the `SceneAssetResolver` FileInfo/ETag/range/symlink contract are untouched. Legacy rows without a prefix keep serving off the filesystem unchanged.

## Changes Made

[...]

PR #2466, opened 2026-07-05
feat(controlplane): substrate-neutral executors - YARP rolling deploy + local process pool

mikemcdougall:
Closes #2460. Related to #2457 (ADR-0060 WS1).

## Summary

Adds the two substrate-neutral control-plane executors that break cloud-lock and deliver the on-prem/air-gapped zero-downtime upgrade story, as peers to the existing ECS/Lambda/ContainerApps/Functions/Batch/K8s backends.

`YarpRollingDeployBackend` (`honua-yarp-rolling`, new `DeployTargetKind.SelfHostedRolling`) implements the Honua-owned rolling cutover on a single host: the server fronts traffic with embedded YARP (config-gated on `ControlPlane:SelfHosted:Enabled`; default deployments never wire the proxy), starts a standby container of the desired revision via the docker/podman CLI, health-gates it against `/healthz/ready`, atomically swaps the proxy destination at promote, and keeps the old replica serving until the new one is confirmed healthy. Rollback handles both pre-cutover (stop standby; old replica untouched) and post-cutover (repoint proxy, drain and stop the failed revision, relaunch the previous revision if the old replica is gone, `ManualInterventionRequired` when unrecoverable). The backend holds no in-memory state: observation reconstructs standby/active from container labels, so the operation survives the server process itself being replaced mid-deploy. Blue/green is deliberately not the primitive — this is statelessness + health-gated rolling replace + a thin executor, per ADR-0060.

`LocalProcessPoolBatchComputeBackend` (`honua-local-process`, new `BatchComputeTargetKind.LocalProcess`) is a true process-spawning GP executor for single-host/air-gapped deployments: shell-free child-process launch with the same `HONUA_*` env contract as the cloud batch backends, a bounded concurrency pool with non-blocking admission (saturation defers via a pending marker instead of holding the reconciliation lease), process-tree kill on cancel, and explicit host-restart semantics (untracked launched jobs fail with a clear message so the reconciler's retry machinery takes over).

The health gate reuses the reconciler's telemetry pipeline: `SelfHostedRolling` maps to the synthetic-probe-only `honua-http` preset. The standby probe uses a dedicated local-replica probe seam that deliberately bypasses the outbound SSRF validator — the target is a sibling replica on loopback built from validated numeric ports, exactly what that guard exists to block for untrusted outbound calls (documented in code).

## Changes Made

- Add `DeployTargetKind.SelfHostedRolling` and `BatchComputeTargetKind.LocalProcess` (additive enum values)
[...]

PR #2464, opened 2026-07-05
feat(ops): enforce expand/contract migration safety and job-contract versioning

mikemcdougall:
Closes #2462. Related to #2457 (ADR-0060 WS3).

## Summary

Enforces the expand/contract discipline that makes rolling upgrades safe, in the two places ADR-0060 requires: (a) the database schema gate — the migration runner now classifies every pending script as expand (additive) or contract (potentially backward-incompatible) and fails closed on contract-phase scripts that lack the established `-- honua:compatibility-review reason=<...>` marker, while the deploy preflight probe blocks coordinated deploys whenever any contract-phase migration is pending (contract migrations must run in a dedicated step, never alongside a rolling deploy); and (b) the serving↔GP-worker job contract — an integer contract version now travels from workload configuration through `ExecutionJobSpec` to the worker as `HONUA_CONTRACT_VERSION`, the dispatcher refuses to submit a job whose version exceeds the target backend's `MaxSupportedContractVersion`, and both worker harnesses (dotnet + python) re-check the injected version and fail closed, so a vX server can never hand a vY worker a job it cannot run mid-upgrade.

The classifier is the single source of truth: the regex/comment-stripping semantics previously inlined in `DatabaseMigrationSafetyTests` were lifted into `MigrationSafetyClassifier` (Honua.Core) and the architecture test now delegates to it. Enforcement defaults ON (`Database:MigrationSafety:Enforce=true`) — safe because every existing breaking migration already carries the marker.

## Changes Made

- Add `MigrationSafetyClassifier` + `MigrationScriptClassification` to Honua.Core; `DatabaseMigrationSafetyTests` delegates to it (one source of truth)
- Extend `DatabaseMigrationPlan` with per-script classifications and contract flags; `PostgresDatabaseMigrationRunner.PlanMigrationsAsync` populates them from script contents
- `PostgresDatabaseMigrationRunner.RunMigrationsAsync` fails closed on unannotated contract-phase scripts, gated by new `MigrationSafetyOptions` (`Database:MigrationSafety`, `Enforce` default true)
- `DeployPreflightProbe` surfaces pending contract scripts and blocks coordinated-deploy readiness while any are pending
- Add job-contract versioning: `CustomCodeJobContract.ContractVersion`/`HONUA_CONTRACT_VERSION`, `ExecutionJobSpec.ContractVersion`, `ExecutionJobDefinition.ContractVersion`, `BatchComputeBackendCapabilities.MaxSupportedContractVersion` (all default 1)
[...]

PR #2501, opened 2026-07-05
fix(ops): correctness+security fixes in migration gate, process pool, YARP deploy, scene hydrator

mikemcdougall:
Related to #2464, #2466, #2467 (correctness/security review of the ADR-0060 operability core).

## Summary

Fixes seven verified defects a fresh review found in the merged operability core — three SEV1, each with a regression test proven to fail before the fix.

**SEV1 — migration-safety gate bypass.** `MigrationSafetyClassifier` stripped `'…'` string literals *before* `--` line comments, so apostrophes in adjacent comments (routine English) opened a quote span that swallowed real DDL between them — hiding a `DROP TABLE` so an unannotated contract migration classified as Expand and passed both the runner fail-closed gate and the deploy preflight block. Replaced the order-sensitive regex passes with a single left-to-right lexer honoring real SQL precedence; added an `ALTER TABLE … RENAME TO` rule. All shipped migrations classify identically (038 stays contract-annotated).

**SEV1 — local process pool burned every retry.** Terminal executions were never evicted from the tracking dictionary, so a reconciler requeue of the same OperationId re-entered `Launch` → "already tracked" → bogus launch failure; in-host retries never relaunched, and the dictionary + scratch dirs grew unbounded. Now evicts after the terminal status is observed once (and deletes the backend-owned scratch dir); a later poll echoes the persisted terminal status so it never flips Succeeded→lost. Fixed the Dispose/monitor `_slots.Release` race.

**SEV1 — YARP rolling deploy blackholed traffic after a restart.** The promoted proxy destination lived only in memory; after a promote (which deletes the old active container) a front-process restart reset the proxy to the now-dead `ActivePort`, and rollback misclassified promoted/not-promoted. Added label/ground-truth reconstruction: a startup reconciliation service repoints the swapper from live container state, and rollback derives `alreadyPromoted` from container ground truth (surviving a restart) — honoring the stateless-projection design.

**SEV2** — scene hydrator manifest path-traversal guard (reject rooted/`..` before any write); scene hydration no longer holds an Npgsql connection during the multi-GB download (materialize row, dispose, then hydrate); `InstallAtomically` now move-aside/move-in/delete rather than delete-then-move (no no-tree window / no 404s mid-republish).

**MINOR** — `HONUA_CONTRACT_VERSION` stamped after the `env.*` passthrough (K8s/AWS) so a workload param can't override the gate value; added to the local backend for consistency.

## Breaking Changes

None.

## Gate Impact

- [x] Strengthens the expand/contract safety gate (previously bypassable)

## Testing

- [x] New regression tests per finding (proven failing pre-fix for the classifier); Core classifier 24/24, arch DatabaseMigrationSafety 8/8, control-plane + scene unit suites 147/147
- [x] Release build (warnings-as-errors) green; format clean


mikemcdougall:

mikemcdougall:

mikemcdougall:

mikemcdougall:

mikemcdougall:

PR #2672, opened 2026-07-11
feat(gp): opt-in local-process custom-code execution backend (OS-sandboxed subprocess) — SECURITY REVIEW REQUIRED

mikemcdougall:
## ⚠️ Security review required — this PR adds an in-process untrusted-code execution surface

This adds a **brand-new code-execution backend** that runs operator-allowlisted, git-pinned custom-code as an **OS-sandboxed subprocess on the honua-server host itself**. It executes untrusted user code in-process (not in a cloud-managed container). **Please give this a human security review, not only automated review, before merge — and again before it is enabled in any real deployment.** It is off by default and fails closed, but the surface is genuinely high-risk.

**This PR has already been through two rounds of adversarial security review.** Round 1 found two blocking issues (no privilege drop; the raw job token was exposed to user code) and three lower-severity gaps — all fixed. Round 2 re-verified round 1's fixes (confirmed the token scrub and the ulimit fail-closed behavior are clean, and empirically verified the UID-drop mechanics genuinely hold), then found one further blocking gap in the UID-drop hardening itself, plus corrected a wrong claim in round 1's own documentation and two undisclosed items. All are fixed/corrected below with an honest before/after — read "Adversarial review findings" before the rest.

## Round 2 findings — before/after (most recent, read first)

| # | Finding | Severity | Before | After |
|---|---|---|---|---|
| **Capability hardening** | `setpriv --reuid/--regid --clear-groups --no-new-privs` does not clear ambient/inheritable capabilities or the bounding set. Walking the actual deployment matrix for granting honua-server `CAP_SETUID`: root → safe (kernel clears caps on `reuid`); non-root + `setcap` file capability → caps don't survive the wrapper's own `execve`, so `setpriv` fails closed (safe, just inert); **non-root + systemd `AmbientCapabilities=CAP_SETUID`** (the exact non-root pattern these docs recommended) → ambient capabilities **do** survive `execve` and are cleared by neither a non-root→non-root `reuid` nor `--no-new-privs` alone, so the "sandboxed" script inherits `CAP_SETUID` and can `setuid(0)` back to root. The only *working* non-root configuration was simultaneously the *escapable* one. | **CRITICAL — blocking** | `setpriv` invocation lacked capability-stripping flags; the ambient-capability escape was undisclosed. | Added `--ambient-caps=-all --bounding-set=-all` to the `setpriv` invocation (`SandboxedProcess.cs`), stripping ambient/inheritable capabilities from the child regardless of how the parent acquired `CAP_SETUID`. Verified with a new white-box test asserting both flags are present in the generated wrapper script whenever `SandboxUser` is configured. |
| **Yama/`ptrace_scope` claim was wrong** | My round-1 docs and code comments claimed Linux's Yama LSM in restricted mode (`ptrace_scope=1`, this repo's own hosts' default) blocks a same-UID child from reading its parent's `/proc/<pid>/environ`. **That's a mischaracterization** — Yama's ancestor-relationship restriction gates only `PTRACE_MODE_ATTACH` (actual attach/trace), not the `PTRACE_MODE_READ_FSCREDS` check `/proc/PID/environ` reads actually use, which is governed purely by same-UID DAC + target dumpability, neither of which Yama restricts. Round-2 review proved this directly: a same-UID child on a `ptrace_scope=1` host successfully read its parent's full environ including a planted secret. | **Documentation correction** | ADR, code doc comments, and a test comment all stated or implied the `/proc/environ` read was conditional on a permissive `ptrace_scope`. | Corrected everywhere (ADR table + prose, `CustomCodeLocalBackendOptions.cs`, `LocalProcessCustomCodeBackend.cs`, and the test file's explanatory comment) to state plainly: this read is **unconditional** on Linux, not mitigated by `ptrace_scope` on any host. No test code change was needed — the existing same-UID-file-read test was already a valid, deterministic demonstration of the underlying risk; only the surrounding prose explaining *why* that vector was chosen was wrong and is now corrected. |
| **Fork bomb unbounded in "acknowledged unconfined" mode** | `MaxProcessCount`/`ulimit -u` correctly does not apply without `SandboxUser` (RLIMIT_NPROC needs a distinct UID — see F3 below), but this meant the F3 narrative read as if the fork-bomb bound applied generally, without stating it does **not** apply in the exact mode most non-`SandboxUser` operators will actually run in. | Undisclosed residual | Not stated explicitly anywhere. | Disclosed explicitly in code docs (`CustomCodeLocalBackendOptions.MaxProcessCount`/`AcknowledgeUnconfinedExecutionRisk` remarks), the backend's class docs, and the ADR: in the acknowledged-unconfined mode, a fork bomb is **not bounded by this backend at all**. |
| **`CAP_SETUID` blast-radius tradeoff** | Requiring `CAP_SETUID` on honua-server is root-equivalent: a honua-server process compromised via **any unrelated vulnerability** can now escalate to root, whereas a plain non-root honua-server could not. This compounds with the capability-hardening finding above (the only working non-root grant path was also the escapable one). | Undisclosed tradeoff | Not stated anywhere. | Added an explicit "blast-radius tradeoff" section to the ADR and a `<remarks>` block on `CustomCodeLocalBackendOptions.SandboxUser` walking the three ways to grant `CAP_SETUID` (root / file-capability wrapper / ambient) and their respective safety and functionality tradeoffs, so an operator configuring this cannot miss it. |

## Round 1 findings — before/after

| # | Finding | Severity | Before | After |
|---|---|---|---|---|
| **F1** | No privilege drop; child ran as the same OS user as honua-server. A same-UID script can unconditionally read any file honua-server can read (DAC) and honua-server's full process environment via `/proc/<pid>/environ` (see the Yama correction above — this is unconditional, not host-dependent), defeating the environment allowlist entirely, plus signal honua-server. | **CRITICAL — blocking** | The "Hard — no host secret is ever inherited" claim was false on a bare host; this gap was undisclosed. | Added `Local:SandboxUser`: the child is switched to a distinct, unprivileged OS user via `setpriv --reuid/--regid --clear-groups --no-new-privs --ambient-caps=-all --bounding-set=-all` (capability flags added in round 2 — see above) before its target program runs; drop failure fails the launch closed. Without `SandboxUser`, the operator must explicitly set `Local:AcknowledgeUnconfinedExecutionRisk=true` — a **code-enforced gate checked on every job**, not just at startup/in docs — or the backend refuses to run anything. |
| **F2** | Raw `HONUA_JOB_TOKEN` handed to user code in cleartext. The sibling **cloud** custom-code path (`docker/worker-customcode-python/harness/.../sandbox.py`) deliberately scrubs this token from the environment before user code runs; this new local path did the opposite and copied it in. | **HIGH — blocking** | `BuildEnvironment` copied `env.HONUA_JOB_TOKEN` into the child's environment. | The token is now **never placed in the child's environment at all** (stronger than "inject then scrub"). Added a defense-in-depth `os.environ.pop('HONUA_JOB_TOKEN', None)` in the Python bootstrap. Documented: this MVP local backend gives no local custom tool a Honua API callback client — use the Batch backend for that. Round 2 confirmed this fix is clean. |
| **F3** | No `ulimit -u` (process count) or `ulimit -f` (output size), so a fork bomb / disk-fill wasn't bounded at all. | Should-fix | Undisclosed DoS surface. | Added `Local:MaxOutputFileBytes` (`ulimit -f`, always) and `Local:MaxProcessCount` (`ulimit -u`, **only** alongside `SandboxUser` — RLIMIT_NPROC is per-real-UID host-wide on Linux). Round 2 added the explicit disclosure that this leaves fork bombs unbounded in acknowledged-unconfined mode (see above). |
| **F4** | The `ulimit` wrapper failed OPEN: `ulimit -v N 2>/dev/null; ulimit -t N 2>/dev/null; exec "$0" "$@"` swallowed a `ulimit` failure and exec'd the target anyway. | Should-fix | Silent fail-open. | Every `ulimit` call now aborts the launch on failure (`\|\| exit 97`). Verified with a white-box test inspecting the generated wrapper script. Round 2 confirmed this fix is clean. |
| **F5** | RLIMIT_CPU counts CPU-*seconds*, not wall-clock — a mostly-sleeping double-forked/detached process can survive the wall-clock kill, the tree-kill, **and a honua-server restart**. | Should-fix / disclose | Disclosed only the double-fork escape, not the persistence detail. | No code fix possible without a container/PID-namespace boundary; explicitly documented as a **persistent-implant risk**, not just a leaked pool slot. |
| — | `SandboxPaths.ResolveContained` uses `Path.GetFullPath` (lexical), not `realpath` — a pinned repo shipping a symlink could point outside the scratch dir. | Lower priority (subsumed by F1) | Undisclosed. | Documented as subsumed by UID separation; no code fix in this MVP, by design per the review's own priority guidance. |

**Test-quality notes addressed across both rounds:** added tests that (a) prove the same-UID risk with a deterministic DAC file-read (not a `/proc/environ` read — not because that read would be blocked, see the Yama correction above, but because the file-read needed no reasoning about which `/proc` file uses which ptrace mode), (b) prove the code-enforced gate rejects a misconfigured backend before any job runs, (c) prove privilege-drop failure fails the job closed rather than running it unconfined, (d) prove `HONUA_JOB_TOKEN` never reaches the child, (e) white-box-inspect the wrapper script for the fail-closed `ulimit` pattern, (f) prove `ulimit -u` is skipped without `SandboxUser`, and (g, round 2) white-box-inspect the wrapper script for the `--ambient-caps=-all --bounding-set=-all` capability-stripping flags.

## Security model (full picture, post-review)

`LocalProcessCustomCodeBackend` is an opt-in `IBatchComputeBackend` selected by `Geoprocessing:CustomCode:Backend=Local` (default `Batch` — unchanged cloud behavior). Six controls, in dependency order:

| # | Control | Mechanism | Strength / honest caveat |
|---|---------|-----------|--------------------------|
| 1 | **Process UID separation** (`Local:SandboxUser`) | `setpriv --reuid/--regid --clear-groups --no-new-privs --ambient-caps=-all --bounding-set=-all` before the target program runs; requires `CAP_SETUID`; drop failure fails the launch closed. | **Hard on POSIX when configured**, including against the ambient-capability escape. The control the others depend on — without it, a same-UID script can unconditionally read any file honua-server can read AND honua-server's full process environment via `/proc` (not mitigated by `ptrace_scope` on any host), and cannot spawn-bomb-bounded processes either (see #4). A code-enforced gate (`AcknowledgeUnconfinedExecutionRisk`) must be explicitly set or the backend refuses to run. Holding `CAP_SETUID` at all is a root-equivalent blast-radius tradeoff — see below. |
| 2 | **Environment allowlist** | Child env is built as an **allowlist** — cleared, then only `CUSTOMCODE_*` contract vars, job-scoped `HONUA_BASE_URL` (never `HONUA_JOB_TOKEN`), a controlled `PATH`, standard `HONUA_*` markers, and operator-named host vars. | Blocks inheritance via the environment vector unconditionally. A real confidentiality boundary only combined with #1. |
| 3 | **Wall-clock timeout + process-tree kill** | Monitor hard-kills the whole tree at `MaxWallClock`. | Hard for the tracked tree; a double-forked escapee can survive it and (since RLIMIT_CPU counts CPU-seconds not wall-clock) persist across a honua-server restart — a persistent-implant risk. |
| 4 | **CPU + address-space + output-size + process-count limits** | POSIX `ulimit -t`/`-v`/`-f`/`-u` via a kernel-enforced, fail-closed launch wrapper; `-u` (`MaxProcessCount`) only alongside `SandboxUser`. | Hard on POSIX, fail-closed, for CPU/AS/FSIZE always. **In acknowledged-unconfined mode, fork bombs are not bounded at all** (disclosed residual). Not enforced in-process on non-POSIX — run in a cgroup-constrained container there. |
| 5 | **Single-use scratch + path-traversal safety** | Fresh per-job dir under `WorkingRoot`, only path handed to the child, deleted on terminal; every constructed path validated to resolve strictly under that root. | Hard for backend-constructed paths; lexical, not symlink-resolving — subsumed by #1. |
| 6 | **Network denial** | *Not enforced by this backend.* | Deployment requirement — run inside an already-network-restricted container/namespace. The one intentional network op (the `git` checkout of the pinned commit) runs in the honua-server process, not the sandbox. |

**The `CAP_SETUID` blast-radius tradeoff (disclosed):** `CAP_SETUID` is root-equivalent. Granting it to honua-server so this backend can drop privileges means any unrelated honua-server vulnerability can now escalate to root. There is no way to get automatic UID-drop-before-exec without the parent holding `CAP_SETUID` in some form. The three ways to grant it have different tradeoffs: **root** (safe for this mechanism, but the whole process is already root), **`setcap` file-capability wrapper** (safe but inert — every job fails closed, doesn't actually work), **systemd `AmbientCapabilities=CAP_SETUID`** (the only non-root option that works — now hardened against the ambient-capability escape by the `--ambient-caps=-all --bounding-set=-all` flags, but the underlying `CAP_SETUID`-on-honua-server tradeoff remains). See `CustomCodeLocalBackendOptions.SandboxUser`'s XML doc and the ADR addendum for the full walk-through.

**Scratch-directory design note (honest):** the single-use scratch reuses the established single-host **local-process scratch pattern** (fresh per-job dir under a configured `WorkingRoot`, deleted on terminal — the same lifecycle shape as the sibling `LocalProcessPoolBatchComputeBackend`), *not* the durable `WorkspaceLifecycleService`/`WorkspaceOptions` machinery, which models durable, TTL'd, owner-scoped GP output artifacts and has no filesystem-root concept. Reviewers should confirm they agree with this boundary.

The upstream submit-time trust gate is **unchanged**: repo allowlist / per-tenant allowlist / signed-commit posture (`CustomCodeRepoPolicy`), full 40-hex SHA pin, scope-⊆-owner clamp, and scoped-token mint all still run. This PR only adds a *place to run* the already-validated code.

## Default-off / additive guarantees

- `Geoprocessing:CustomCode:Backend` defaults to `Batch` → the local backend, preparer, and catalog decoration are **not even registered**. A deployment that hasn't opted in is byte-for-byte unchanged.
- Even when `Backend=Local` is selected, the backend **fails closed** at `StartAsync` unless `Geoprocessing:CustomCode:Local:Enabled=true`, **and** unless either `SandboxUser` is configured or `AcknowledgeUnconfinedExecutionRisk=true` is explicitly set (both re-checked on every job, not just at startup).
- `CustomCodeOptions.RepoPolicy` default is unchanged (fail-closed empty allowlist), so selecting Local alone does not enable execution.

## What's here

- `LocalProcessCustomCodeBackend` — the sandboxed-subprocess `IBatchComputeBackend` (`honua-local-customcode`, `TargetKind=LocalProcess`), modeled on the proven `LocalProcessPoolBatchComputeBackend` with the six controls above.
- `SandboxedProcess` — the security core: env-clear + allowlist, POSIX `ulimit` wrapper (fail-closed) + `setpriv` UID-drop with capability stripping, `SandboxPaths` containment/sanitization helpers.
- `ICustomCodeWorkloadPreparer` + `GitPythonWorkloadPreparer` — the git-checkout + python-launch seam. Python bootstrap defense-in-depth-scrubs `HONUA_JOB_TOKEN`.
- `CustomCodeLocalBackendOptions` (+ validator) — all knobs default to the safe/off posture; the validator fails startup if `Enabled=true` without either `SandboxUser` or the explicit acknowledgement.
- Discovery: `CustomCodeLocalToolCatalog` decorates `IProcessCatalog` so operator-registered local tools surface through the GPServer task-listing / task-info routes with real parameter-schema metadata.
- DI wiring in `AddGeoprocessing` (gated on `Backend=Local`); the local-process backend registration in `BatchAndDeployBackendsRegistration.cs` is untouched.
- ADR-0057 addendum documenting the isolation boundaries and their limits, corrected across both review rounds.

## Tests (adversarial)

`LocalProcessCustomCodeBackendTests` (24 tests) — each control has a test that tries to break it (POSIX `/bin/sh`, mirroring `LocalProcessPoolBatchComputeBackendTests`; runs on the Linux CI host):

- **UID separation**: the code-enforced gate rejects a misconfigured backend before any job runs; a same-UID DAC file read succeeds when unconfined (documents the residual risk); a `SandboxUser` whose privilege drop fails (non-root CI) fails the job closed and the target script never runs; a white-box test confirms `--ambient-caps=-all --bounding-set=-all` are present in the generated script whenever `SandboxUser` is configured.
- **Token exposure**: `HONUA_JOB_TOKEN` never reaches the child even when the job spec carries it.
- **Env**: a host secret set on the parent is **not** leaked to the child; an allowlisted var and a `CUSTOMCODE_*` contract var **are** present.
- **Wall-clock**: a `while true` job is killed and reported `Failed` with a "wall-clock" message.
- **Limits**: `ulimit -t`/`-v`/`-f` read back inside the sandbox equal the configured ceilings; `ulimit -u` is skipped without `SandboxUser`; a memory-hog is aborted by RLIMIT_AS; a white-box test confirms the wrapper script's `ulimit` calls all fail-closed with no swallow.
- **Scratch/traversal**: the child's cwd is inside the scratch dir and deleted after terminal; `SandboxPaths` rejects `..` and absolute escapes; a `deps_manifest` traversal is rejected closed; a malicious `SandboxUser` value is rejected even if the options validator was bypassed.
- **Fail-closed**: disabled backend rejects; discovery surfaces a configured tool with its schema; options-validator tests cover the `SandboxUser`/acknowledgement gate, `root` rejection, and charset rejection.

## Known MVP limitations (documented, not hidden)

- **Python only**; `.NET` custom code fails closed with a clear message.
- **No dependency installation** (the sandbox has no network); python jobs must be dependency-free or run against a pre-provisioned interpreter.
- **No Honua API callback client** for local custom tools in this MVP — use the Batch backend if a tool needs to call back into honua-server.
- **Filesystem/network hardening** beyond the in-process controls needs a surrounding container/namespace.
- `Local:SandboxUser` requires honua-server to hold `CAP_SETUID` — an operational prerequisite with the blast-radius tradeoff disclosed above, not something this code can bootstrap itself.
- **Fork bombs are unbounded in acknowledged-unconfined mode** (no `SandboxUser`) — disclosed, not fixed in this MVP (fixing it would require RLIMIT_NPROC against a UID that isn't shared with honua-server, i.e. `SandboxUser`).

## Verify

`dotnet build` (Honua.Geoprocessing) and `dotnet test` (Honua.Server.Tests, filtered to `LocalProcessCustomCodeBackendTests`, 24 tests) green on Linux (WSL; this repo's CI target for POSIX-subprocess tests, matching the sibling `LocalProcessPoolBatchComputeBackendTests`).

This has been through two rounds of adversarial review and all findings are fixed or explicitly disclosed. **Per the coordinator, this is intended to be the final automated round — it now goes to the user for the human security sign-off this PR has always required, regardless of how clean the automated findings are.**


mikemcdougall:
Merge-blocking security review:

- This is not an OS sandbox: untrusted code runs as the Honua server UID with access to host/container files, mounted secrets, Unix sockets, network targets, and cloud metadata. `AllowNetwork=false` is advisory and unenforced.
- Resource controls fail open because failed `ulimit` calls are ignored. There is no PID/process-count, disk/file quota, namespace/cgroup boundary, or reliable containment of reparented descendants.
- Custom process IDs are not reserved against the conventional aliases introduced by #2675, so a custom `Buffer` process can hijack standard routing.
- Discovery metadata is not bound to repository/SHA/entrypoint and can advertise a schema that omits required trust inputs.

Default-off is useful but insufficient for an execution boundary. Require an enforceable container/namespace/cgroup profile (and refuse startup when it cannot be verified), or explicitly scope this to trusted operator code without calling it sandboxed. Add the required issue/template and full governance gates before returning to the train.

mikemcdougall:
Re-review of current head cb46628be: still blocked.\n\n- Default Linux UID-separated launch emits ulimit -u under /bin/sh; Ubuntu dash rejects that option, so the documented default path exits before setpriv. The test only inspects the command string.\n- The documented CAP_SETUID grant is insufficient: dropping the bounding set needs CAP_SETPCAP, and reliably killing a different-UID child needs CAP_KILL. TryKill swallows permission failures and then waits without cancellation, invalidating the wall-clock/slot guarantees.\n- Windows accepts SandboxUser as sufficient configuration but launches directly without UID separation or resource limits. Non-POSIX must reject this mode or require a verifiable external sandbox.\n- Cancellation during preparation can leak a slot and scratch directory; post-launch cancellation can leave execution/scratch cleanup dependent on a later observation.\n- Duplicate configured process IDs and built-in alias collisions are silently overwritten/skipped rather than rejected at startup.\n\nToken scrubbing and explicit opt-in defaults are improvements. A successful privilege-separated execution is still not tested, which is why the launch and capability-contract failures remain invisible.

mikemcdougall:
Closing: **no-go by decision.** Custom-code execution must remain **AWS-Batch-only** (isolated, per-job container), which is the existing default (`Backend=Batch`). This PR's local-process-on-host backend (`Backend=Local`) runs operator-supplied untrusted code as a subprocess of honua-server itself — outside a cloud-managed container — and its only realistic isolation depends on deployment-specific container boundaries plus operator-configured privilege drop. That posture is not one we want to offer, even opt-in/disabled-by-default.

Follow-up (separate PR): an ADR codifying "custom code execution is GP AWS-Batch-only," plus enforcement guards (a) server-side rejecting any non-Batch custom-code backend selection, (b) in the Batch custom-code Docker harness, and (c) in the SDK — so this can't be reintroduced by config or client.

The adversarial security review work here (privilege-drop, capability stripping, token scrubbing, `/proc/environ` disclosure correction) remains valuable reference for hardening the sanctioned Batch path; nothing is lost by keeping it in the PR history.

Issue #2765, opened 2026-07-11
GP worker: untrusted base64 raster opened by GDAL with format auto-detect — VRT SourceFilename / /vsicurl file-read + SSRF (suspected)


mikemcdougall:
## Suspected finding (source-traced, not runtime-confirmed — no live server in review env)

Raster GP inputs arrive as **base64 blobs opened by GDAL with format auto-detection** (`GdalCalcInputs.cs`, `GdalJobInputReader.cs`). A caller can base64-encode a GDAL **VRT** (an XML format) whose `<SourceFilename>` points at an arbitrary local path (`/etc/passwd`, a mounted secret) or a `/vsicurl/…` / `/vsis3/…` URL — a known GDAL **file-read / SSRF** vector — unless the worker restricts the driver set (`GDAL_SKIP` / an explicit allowed-driver allow-list) and disables the `/vsicurl`-family handlers.

No explicit driver allow-list was found in `src/Honua.Worker.Gdal/`. GDAL will happily open a VRT presented as a "GeoTIFF" because it sniffs content, not the claimed media type.

## Blast radius
Contained to the custom-code/GP worker container, which per ADR-0063 runs isolated (non-root, credential-scrubbed) in AWS Batch — so impact is **in-container file read + limited SSRF from the worker's network position**, NOT host compromise. Hence Medium, not Critical. But the worker's network position and any mounted config/secrets are still meaningful.

Note: this is distinct from classic XXE — it's GDAL VRT `SourceFilename` resolution / `/vsicurl`, not general XML entity expansion.

## Ask
1. Verify whether the worker restricts GDAL drivers (allow-list to GTiff/PNG/JPEG/the intended raster formats; exclude VRT and other indirection drivers) and disables `/vsicurl`/`/vsis3`/`/vsihttp` for untrusted inputs (e.g. `GDAL_HTTP_*` / `CPL_VSIL_CURL_*` / `GDAL_SKIP=VRT` or an inclusion list via `GDAL_DRIVER_PATH`/open-options).
2. If not restricted, add a driver allow-list + disable remote VSI handlers on the untrusted-input open path.

## Source
Found in an adversarial GP security review. Worker input path: `src/Honua.Worker.Gdal/Execution/{GdalCalcInputs.cs,GdalJobInputReader.cs}`; runners `ProcessGdalCommandRunner.cs`.

PR #2780, opened 2026-07-12
fix(worker-gdal): pixel-dimension admission control + zonal zone bound (#2766)

### LINKED ISSUE #2766: GP worker: no pixel-dimension admission control → decompression-bomb OOM; unbounded zonal-statistics zone count
## Finding (source-traced)

The raster worker caps aggregate **base64/decoded byte size** (`MaxArtifactBytes`, `GdalCalcInputs.cs:104-105`, with a pre-decode encoded-length precheck in `GdalJobInputReader.cs:133-165`) — but there is **no bound on decoded pixel dimensions** (width × height × bands × dtype). A few-KB, highly-compressible GeoTIFF (or one declaring enormous width/height) forces the worker to allocate width×height×bands×dtype when GDAL processes it → **decompression-bomb OOM**.

No `MaxPixels` / `MaxWidth` / `MaxHeight` / max-band admission control was found in `src/Honua.Worker.Gdal/`.

Secondary, related (same file area): **unbounded zone count** in zonal-statistics — `GdalRasterZonalStatisticsJobExecutor.cs:209-212` applies a per-invocation timeout but the cumulative per-zone work is unbounded, so a payload with a huge zone FeatureCollection can exhaust the worker.

## Blast radius
Authenticated caller (has `Process.Execute`); contained to the isolated worker container, bounded only by whatever memory/CPU limit the Batch/K8s job sets. Medium — a resource-DoS on the GP compute tier, not host compromise or data disclosure. (Admission control `ExecutionAdmissionEvaluator` bounds job COUNT/cost-weight/rate, not per-job pixel footprint.)

## Ask
Add pixel-dimension / cell-count admission control on raster inputs (reject before GDAL allocates: cap total pixels, max width/height, max bands), and bound the zone count (or total zone vertices) in zonal-statistics. Consider tying the cost-weight in `ExecutionAdmissionEvaluator` to declared/derived pixel count so oversized rasters are rejected/deprioritized at admission.

## Source
Adversarial GP security review + GDAL-worker injection sub-audit (both independently flagged the pixel-dimension gap). Files: `src/Honua.Worker.Gdal/Execution/{GdalCalcInputs.cs, GdalRasterZonalStatisticsJobExecutor.cs}`, `ExecutionAdmissionEvaluator.cs`.

mikemcdougall:
Fixes #2766

## Problem

The GDAL raster worker capped aggregate decoded **byte** size (`MaxArtifactBytes`)
but placed **no bound on decoded pixel dimensions** (width × height × bands ×
dtype). A few-KB, highly compressible raster that *declares* enormous
width/height forces GDAL to allocate `width×height×bands×dtype` on open — a
decompression-bomb OOM. Separately, `raster.zonal-statistics` applied a
per-invocation timeout but left the **zone count unbounded**.

## Fix

**Pixel-dimension admission control (`GdalRasterDimensionGuard`).** Before any
temp file is written or GDAL is invoked, the guard reads the raster's **declared**
dimensions straight from the header bytes already present at the front of the
(byte-bounded) decoded blob — no subprocess, no full-raster open, no GDAL
dependency — and rejects when the declared footprint exceeds a cap.

**Formats covered — the accepted-format set is content-sniffed by GDAL, so the
guard covers every compressible container that can be tiny-on-disk yet declare a
huge canvas:**
- **TIFF / BigTIFF** — first-IFD `ImageWidth`/`ImageLength`/`SamplesPerPixel`/
  `BitsPerSample` (classic + BigTIFF, little/big-endian).
- **PNG** — `IHDR` width/height/bit-depth/colour-type (palette/indexed counted as
  RGBA so the estimate does not under-bound GDAL's palette expansion).
- **JPEG** — `SOFn` frame width/height/components (marker scan skips APP0 etc.).

This closes the non-TIFF bypass: because GDAL opens by content-sniffing (not the
`input.tif` name) and there is no downstream driver/`-if` restriction, a PNG whose
`IHDR` declares 100000×100000 (or a JPEG at 65535×65535 ≈ 12.8 GB decoded) is now
rejected up front. For any format the guard still can't parse it **admits** and
lets GDAL adjudicate — the residual is a format that is neither TIFF, PNG, nor
JPEG; this is complementary to the driver-allowlist hardening in the sibling PR
#2765 (that PR bounds *which* drivers open; this PR bounds *dimensions* for every
format admitted). A BigTIFF `LONG8` dimension past `Int64.Max` **fails closed**
(clamped to a positive absurd value the caps reject) rather than wrapping negative
and being silently admitted.

**Zone bound (`GdalZoneAdmission`).** Rejects a `raster.zonal-statistics` payload
whose zone-feature count exceeds `MaxZoneCount` or whose cumulative geometry
vertex count exceeds `MaxZoneVertices`, before the per-zone gdalwarp+gdalinfo loop.

**Configurable caps (`GdalWorkerOptions`, restrictive defaults, data-annotation
validated, bound from the `GdalWorker` config section):**

| Option | Default | Rationale |
|---|---|---|
| `MaxRasterWidth` / `MaxRasterHeight` | 100 000 | bounds a single runaway dimension |
| `MaxRasterPixels` | 500 000 000 (500 MP) | fits a full Sentinel-2 tile (~120 MP) / moderate mosaic; rejects 100k×100k = 10 000 MP |
| `MaxRasterBands` | 512 | hyperspectral tops ~250 bands; rejects thousands |
| `MaxDecodedRasterBytes` | 4 GiB | conservative allocation **floor** (w×h×bands×first-band-dtype); excludes warp/block-cache buffers, dtype promotion, palette/RGBA expansion — real peak can exceed it, so it is a lower-bound gate |
| `MaxZoneCount` | 10 000 | covers county/watershed sets; bounds subprocess fan-out |
| `MaxZoneVertices` | 2 000 000 | bounds cutline/geometry cost independent of zone count |

Wired into all 17 raster-source executors (calc family, mosaic — incl. its private
decoder, clip, reproject, reproject-catalog, resample, format-convert, contour,
polygonize, proximity, surface, viewshed, statistics/histogram, zonal-statistics)
at each source-decode point, so the reject precedes the expensive op.

Item 4 (tying `ExecutionAdmissionEvaluator` cost-weight to declared pixel count)
was **not** done: that evaluator runs at control-plane admission without the
decoded raster bytes, so bounding per-job pixels there is not a clean addition —
the worker-side header guard is the correct chokepoint.

**Out of #2766 scope (flag only, to be filed separately):** `raster.rasterize`
and `raster.interpolate` (`gdal_grid`) accept an attacker-controlled **output**
grid size (`-outsize`/`-ts`/`-tr`) that is unbounded — a distinct
output-allocation DoS this guard (an *input*-dimension bound) does not address.

## Tests (no GDAL binary required)

- `GdalRasterDimensionGuardTests` — parses classic LE/BE + BigTIFF, PNG, and JPEG
  headers; each cap (pixels / width / height / bands / decoded-bytes) rejects; a
  ~60-byte TIFF, a PNG, and a JPEG declaring huge dimensions are each rejected from
  the header alone; a BigTIFF width > Int64.Max rejects (fail-closed); unrecognized
  bytes are admitted.
- `GdalZoneAdmissionTests` — over-count and over-vertex rejected; normal sets
  admitted.
- `GdalRasterAdmissionExecutorTests` — at the executor level, an over-pixel-cap and
  over-band-cap TIFF **and a PNG bomb** are rejected with `runner.Invocations`
  empty (proving the reject precedes any GDAL spawn/allocation); a zonal payload
  over the zone cap is rejected before the per-zone loop; a normal in-bounds raster
  and a normal 2-zone set still succeed end-to-end.

**Needs a real gdal binary (not covered here):** confirming that a genuine large
GeoTIFF/PNG/JPEG opened by the actual GDAL CLIs is admitted/rejected identically —
the existing `GdalCli`-gated integration suites cover the real-binary path when
GDAL is present.

## Build / test

`dotnet build` clean; `dotnet test Honua.Worker.Gdal.Tests` → **201 passed, 0
failed, 3 skipped** (the 3 skipped are the gdal-binary-gated integration tests).


mikemcdougall:

Issue #2793, opened 2026-07-12
GP worker: bound caller-controlled output resolution (-tr/cellSize) in resample and rasterize


mikemcdougall:
## Problem Summary

Follow-up to #2782 (merged output-grid caps for explicit `-ts`/`-outsize` paths). The caller-controlled **resolution** paths remain unbounded, confirmed during the adversarial review of the #2782 fix:

- `GdalRasterResampleJobExecutor` — `gdalwarp -tr cellSizeX cellSizeY`: cellSize is validated positive+finite only; output pixels = extent ÷ cellSize, so a tiny cellSize over a wide-extent input yields billions of output pixels. `GdalOutputGridGuard` is never consulted on this branch.
- `GdalRasterizeJobExecutor` — `cellSize` → `-tr` branch: same class; extent comes from the caller-supplied vector input.

Only `ToolTimeout`/`MaxArtifactBytes` backstop these paths today (a worker burns a full timeout or OOMs per request). The residual is documented in `GdalOutputGridGuard`'s XML remarks.

## Why This Matters Now

Authenticated resource-DoS on the GP compute tier survives the #2782/#2784 hardening; the guard covers only the explicit-grid branch of the same executors.

## Acceptance Criteria

- [ ] Output pixel count derived from (input extent ÷ cellSize) is bounded pre-spawn for the resample `-tr` path (input raster extent is available from the dimension-guard/gdalinfo path or a cheap header read).
- [ ] The rasterize `cellSize` → `-tr` branch is bounded (extent from the vector payload envelope, or a documented conservative minimum-cellSize/maximum-extent-ratio cap).
- [ ] Over-cap requests rejected with a clear validation error before the subprocess spawns; tests prove pre-spawn rejection.
- [ ] `GdalOutputGridGuard` remarks updated to remove the residual once closed.

## Affected Repo(s)

honua-server

## Gate Tier Impact

Fast tier (Honua.Worker.Gdal.Tests).

## Release/Deploy Impact

None.

## Non-Goals

Bounding GDAL-internal auto-derived resolutions (reproject without -tr) — bounded by input caps already.

Issue #2770, opened 2026-07-11
bug(mcp): honua_render_map returns generic internal error for all layers on the serverless/Lambda demo


mikemcdougall:
## Problem Summary

`honua_render_map` returns a generic `internal` error for **every** layer tried on the live serverless/Lambda demo (`demo.honua.io`, image `nightly-lambda-aot-20260711`, AOT Lambda). Input validation passes (well-formed `{serviceId, layerId}` refs, `bbox`, `bboxSrid`, `width`, `height`) — the failure is at execution/render time, so no image is ever produced. Observed across a vector layer (`maui-buildings` FeatureServer layer 13), a hillshade layer (`maui-hillshade` layer 7), and an imagery layer (`maui-imagery` layer 9). The returned envelope is safe (no internals leaked):

```json
{"status":"error","code":"internal","message":"An internal MCP operation failed.","error":{"kind":"ExecutionFailed","message":"An internal MCP operation failed."}}
```

This is pre-existing and distinct from #2578: the 2026-07-07 cert recorded `render_map` returning a response that failed *output-schema* validation (fixed by #1956 — the envelope now validates against the `oneOf` error branch), but the underlying runtime render failure remains.

## Why This Matters Now

`honua_render_map` is an advertised MCP tool that never succeeds on the serverless/Lambda demo — the primary public showcase. MCP clients get a schema-valid but useless generic error with no rendered map. It also means the render path has no working end-to-end coverage in the AOT/serverless profile, so regressions there are invisible.

## Steps to Reproduce

1. Authenticate to the demo MCP endpoint (`POST https://demo.honua.io/mcp`, header `X-API-Key: <admin key>`), `initialize` → `notifications/initialized`.
2. `tools/call` `honua_render_map` with valid args, e.g. `{"layers":[{"serviceId":"maui-hillshade","layerId":7}],"bbox":[-156.7,20.7,-156.4,20.95],"bboxSrid":4326,"width":256,"height":256}`.
3. Observe `structuredContent` = the generic `internal` / `ExecutionFailed` error; no `imageBase64`.

## Expected Behavior

`render_map` returns a rendered map image (`imageBase64` + `format`/`width`/`height`) for a renderable demo layer — or, if map rendering is intentionally unsupported in the serverless/AOT Lambda profile, a **specific, actionable** capability error (e.g. `failed_precondition` naming the missing capability), not a generic `internal` error.

## Actual Behavior

[...]

Issue #2457, opened 2026-07-05
Epic: Operate slice 2 — ask–explain–fix on the paved path (serving-plane findings, LLM seat, truthful rollback, self-operating demo)


mikemcdougall:
# Operate slice 2: ask–explain–fix on the paved path

**Thesis.** The Operate loop is real on `trunk`: `/api/v1/operate/status`, ops-health + persisted history, ten deterministic findings, propose/approve/reject with the console `/inbox` as the only approval seat, the guardrail ladder and ADR-0062 autonomy policies with a kill switch, the health-gated deploy reconciler with telemetry evaluators over Argo/YARP/ACA/Functions/ECS/Lambda/docker backends, SignalR push, eight MCP ops tools, and console `/operate/copilot|health|deploy|releases`. What it demos honestly today is "findings → propose → approve → health-gated rollback → kill switch". What it cannot yet demo is **"ask why X is slow and one-click fix it"** — there are no serving-plane findings (nothing answers "why is the tile cache slow"), only four executable actions, the GitOps hand-off backends advertise rollback they cannot perform, the Helm chart registers no deploy target, and the demo still rolls back with a manual `aws lambda update-alias`. Slice 2 closes exactly that gap on the paved path — server-owned findings and actuators, an LLM seat that lives in the console via the server-side proxy (never server-side model calls, ADR-0028), truthful rollback everywhere, and a demo that operates itself.

**Foundations are DONE and out of scope here:** WS0–WS4 (#2459–#2463) and every track of the sibling epic #2552 (closed) — dashboard memory/realtime, agent-seat completion, deploy/upgrade safety, self-healing fabric, proof/packaging. Do not re-open them; build on them.

## MVP (2026.1)
| # | Repo | Ticket | Size | Dep |
|---|---|---|---|---|
| O-A | server | #3300 serving-plane findings v3 (per-service latency/error, tile-cache hit-rate & seed backlog, pool saturation, slow-query hotspot) | M | — |
| O-B | server | #2661 tile-cache lifecycle → `tiles.seed` / `tiles.expire` OpsActions (RequiresApproval tier); expire/delete/inventory half is stranded PR #2835, recovered under #3248 | L | #3248, O-A |
| O-C | server | #3301 rollback truthfulness (real revert or `SupportsRollback=false`) + capability keys `ops.findings` / `ops.autonomy` / `deploy.rollback` | M | — |
[...]

PR #2816, opened 2026-07-13
fix: backlog batch — maps collision resolution, GP mutating-tier authz, meter isolation, raster GP parity, WCS Zarr subsetting

### LINKED ISSUE #2698: Raster/surface process ids excluded from FirstSliceAutomatedVectorProcessIds — 404 on direct process-id execution
## Summary

`ProcessMigrationEvidenceClassifier.FirstSliceAutomatedVectorProcessIds` (the allowlist that gates direct `POST/GET /ogc/processes/processes/{processId}/execution` and `.../{processId}` description calls) contains only vector/geometry process ids (buffer, clip, intersect, project, simplify, snap, dissolve, etc.). None of the 19 raster/surface process ids in `BuiltInProcessCatalog.cs` (`surface.slope`, `surface.aspect`, `surface.hillshade`, `surface.contour`, `surface.viewshed`, `surface.roughness`, `surface.rugosity-tpi`, `surface.rugosity-tri`, `raster.clip`, `raster.mosaic`, `raster.reclassify`, `raster.reproject`, `raster.resample`, `raster.map-algebra`, `raster.spectral-index`, `raster.statistics`, `raster.histogram`, `raster.interpolate-idw`, `raster.interpolate-kriging`, `raster.zonal-statistics`) are in it.

## Observed behavior

A direct call like `POST /ogc/processes/processes/surface.slope/execution` returns `404 no-such-process`, even though `surface.slope` is a real, registered `ProcessDefinition` with a working executor (`GdalSurfaceJobExecutor.cs`). The only way to invoke a raster tool today is to wrap it as a single step inside the canonical `honua-geoprocessing` process's `plan` input (confirmed via `tests/dotnet/Honua.Protocols.OgcApi.Tests/Source/Processes/OgcProcessesExecutionSubmissionTests.cs:68`).

## Why this is worth a maintainer look

This may be entirely deliberate — raster/surface tools route to the out-of-process GDAL worker (`src/Honua.Worker.Gdal/Execution/`), and `FirstSliceAutomatedVectorProcessIds`'s naming ("vector", "first slice") suggests it may be intentionally scoped to a synchronous in-process fast path that doesn't (yet) apply to worker-dispatched raster jobs. But nobody currently documents that reasoning, and it creates a real asymmetry: a client author has to know, out of band, that vector process ids are directly callable but raster ones require plan-wrapping — there's no error message or discovery mechanism that explains why `surface.slope` 404s while `geometry.buffer` doesn't.

## Ask

Could someone with context on `ProcessMigrationEvidenceClassifier`'s design confirm whether this exclusion is intentional (and if so, whether the 404 could carry a clearer message pointing callers at the `honua-geoprocessing` plan-wrapping path), or whether raster tools should eventually get first-slice/direct-process-id parity with vector tools?

## Source

Found while building raster-GP support into honua-sdk-python's `honua_sdk.geoprocessing` client — the client works around this by always wrapping raster process calls into a single-step plan, but that workaround shouldn't need to exist if the exclusion isn't deliberate.

### LINKED ISSUE #2796: bug(wcs): Zarr GetCoverage ignores RANGESUBSET and 404s over-extent trims instead of intersecting
## Problem Summary

Found in the post-merge review of PR #2728 (issue #2696). The new Zarr GetCoverage branch (`Wcs20Handler.HandleZarrGetCoverageAsync`) has two contract gaps:

1. **RANGESUBSET silently dropped (wrong data).** `TryResolveCoverageQuery` parses and validates `RANGESUBSET` into `query.Bands`, but the Zarr branch only inspects `OutputFormat`/`OutputSrid`/`ResamplingAlgorithm`. A request like `RANGESUBSET=band2&SUBSET=elevation(500)` on a Zarr-registered multi-band coverage validates, then returns the resolved primary-variable grayscale PNG with 200 — the band selection is ignored. This contradicts #2696's "truthfully advertise/reject unsupported combinations" criterion.
2. **Over-extent spatial trims 404 instead of intersecting.** The handler forwards the raw user envelope; `ZarrRasterSliceReader` requires full containment against the Zarr store's extent → `InvalidSubsetting` 404. WCS 2.0 core semantics (and the plain `IRasterStore` path, exercised by CITE) answer over-extent trims with the intersection. Worse, DescribeCoverage advertises the primary raster's extent while containment is checked against the Zarr metadata extent — float-rounding differences can 404 a client echoing the advertised bbox.

Also: the oversize-slice test (`SCALESIZE=x(4097)`) is intercepted by the pre-existing `TryResolveScaleSize` cap and never reaches the new guard, and the guard's no-scaling-operator path emits `locator="SCALESIZE"` for requests containing no SCALESIZE parameter.

## Why This Matters Now

Silent wrong data on a public WCS surface (finding 1) and a spec divergence on the new cloud-native path (finding 2) shipped without divergence documentation or endpoint tests, contrary to the adapter rule.

## Steps to Reproduce

1. Register a multi-band coverage with a Zarr store and declared axes.
2. `GetCoverage ... &RANGESUBSET=band2&SUBSET=<axis>(<coord>)` → 200 with primary-variable image (band selection ignored).
3. `GetCoverage ... &SUBSET=Long(-190,180)` → 404 InvalidSubsetting (expected: intersection response).

## Acceptance Criteria

- [ ] Zarr branch rejects `query.Bands` with `OperationNotSupported` / locator `RANGESUBSET` (or implements band selection).
- [ ] Spatial trims clamped to the intersection with the coverage extent before the Zarr read; `InvalidSubsetting` only on empty intersection (matching temporal behavior); DescribeCoverage-advertised extent round-trips.
- [ ] Oversize guard covered by a test that actually reaches it (large base grid, no scaling operator) with a truthful locator.
- [ ] Any remaining intentional divergence documented in the parity doc + divergence test per the adapter rule.

## Affected Repo(s)

honua-server

## Gate Tier Impact

PR gates (build, test). No CITE impact (CITE data has no Zarr registrations).

## Release/Deploy Impact

None.

## Non-Goals

Implementing full multi-band Zarr composition; TIFF/JPEG output for Zarr slices (separately rejected by design).

Related to #2696, PR #2728.

### LINKED ISSUE #2798: GP authz: workflow-run surface bypasses the mutating-process tier via the synthesized admin orchestrator principal
## Problem Summary

Found in the review of PR #2773 (merged; issue #2754). The new `ExecuteMutatingProcess` tier is enforced in `GeoprocessingJobService.SubmitJobAsync` and every protocol adapter funnels the caller's principal through it — but the workflow-run surface routes around the gate:

- `OrchestrationSystemPrincipal` submits every workflow step job under a synthesized principal carrying `role=admin`, which hits `OperatorAuthorizationEvaluator`'s unconditional admin bypass.
- `WorkflowOrchestrationEngine.CreateRunAsync` performs no authorization; `WorkflowPackageService.RunPublicationAsync` checks only baseline `OperatorOperation.Execute` before creating the run (its own in-code comment anticipates the endpoint policy being relaxed).

**Mitigation today:** the whole workflow-package endpoint group is `.RequireAdminAuthorization()`, so callers are admins anyway — the bypass is latent, not live.

**Failure scenario:** the moment publication-run opens to non-admin operators (the documented plan), an Execute-only role schedules a publication whose compiled plan contains `sink.honua-layer`/`import.dataset`/`data-management.delete-features`; steps execute under the admin-bypassing orchestrator principal and `ExecuteMutatingProcess` is never evaluated against anyone.

Secondary hardening gaps from the same review:
- `ProcessDefinition.ExecutionTier` defaults to Analytic (fail-open by omission for future mutating built-ins); only the 8 known ids are pinned — no catalog invariant forces a tier decision on new processes. This is also the third hand-maintained side-effect list alongside `ProcessDestructiveClassifier`'s two.
- The custom-code branch skips both `EnsurePlanCatalogValid` and `ContainsMutatingProcess`; nothing asserts custom-code submissions carry no executable Geoprocess steps (inert today).
- Mutating-tier 403s are logged as `Execute` denials (indistinguishable in security logs); no per-protocol end-to-end 403 test.

## Why This Matters Now

#2754 closed with PR #2773, but the tier boundary is one endpoint-policy relaxation away from being bypassable on the workflow surface.

## Acceptance Criteria

- [ ] `ExecuteMutatingProcess` (and custom-code tier) evaluated against the REQUESTING principal at workflow run/publication creation when the compiled definition contains mutating steps.
- [ ] Catalog invariant test forcing an explicit tier decision for every process (snapshot or category assertion for sink/import/data-management).
- [ ] `ContainsMutatingProcess` runs unconditionally (or Geoprocess steps rejected for custom-code submissions).
- [ ] Denial logging carries the actual operation; one end-to-end 403 integration test on a protocol surface.

## Affected Repo(s)

honua-server

## Gate Tier Impact

PR gates (build, test).

## Release/Deploy Impact

None.

## Non-Goals

Removing the evaluator's admin bypass generally; redesigning orchestrator identity.

Related to #2754, PR #2773.

### LINKED ISSUE #2799: bug(ogc-maps): duplicate-binding fix stops at the handler — renderer re-resolves via first-wins index; map/tiles siblings still 404; collisions stay silent
## Problem Summary

Follow-up from the review of PR #2774 (merged; auto-closed #2759). The handler-level resolver now picks the Maps-enabled resource deterministically, but:

1. **Renderer re-resolves through the same first-wins index.** `OgcMapsRenderingHandler` resolves the vector resource, then calls `_mapRenderer.RenderCollectionMapAsync(layerId, …)` with the bare id; `VectorAwareRasterMapRenderer` does its own `ResourcesByStorageLayerId.TryGetValue` — first-wins — so for the colliding layer it gets the raster resource, `HasGeometry` is false, and the styled-vector path is skipped. Depending on raster rows, the endpoint either still 404s ('No map data found…') or serves raster bytes for a request the handler authorized as vector. The PR's regression test mocks `IRasterMapRenderer`, so it proves handler-level resolution only, and no JS compat lane was dispatched from the branch — **#2759's acceptance criterion (MapLibre/OpenLayers lanes pass) is not demonstrated**. If the lanes still fail, reopen #2759 or track it here.
2. **Sibling map-tiles endpoints still 404** for the same collection: `OgcMapsTileSetHandler.ResolveResourceAndServiceAsync` uses the unfixed first-wins pattern, so `…/map` = 200 while `…/map/tiles` = 404 (and the tiles list links back to the working map) — capabilities/runtime consistency violation.
3. **The collision stays silent system-wide.** ~15 other consumers of `ResourcesByStorageLayerId` (feature stores, WCS, permanent filters, streaming subscription filters, …) still first-wins-resolve the colliding id; the fallback path emits no warn log and graph validation raises no soft error for colliding StorageLayerIds, so the next wrong-resource symptom is undiagnosable.
4. Minor: the unfiltered `GET /ogc/maps/map` dataset-map path enumerates the index directly, excluding the colliding collection that the `collections=<id>` path includes.

## Why This Matters Now

The JS compatibility lanes were the reporting surface for #2759; without renderer-level resolution the same 404 (or a wrong-dataset render) plausibly persists post-merge.

## Steps to Reproduce

Seed two storage bindings sharing a StorageLayerId (raster-first), then: `GET /ogc/maps/collections/{id}/map` with real renderers (no mocks) → 404/wrong bytes; `GET /ogc/maps/collections/{id}/map/tiles` → 404.

## Acceptance Criteria

- [ ] Resolved resource identity flows from the handler into the renderer (no second first-wins lookup); real-renderer integration test over a duplicate-binding fixture.
- [ ] Map-tiles handlers share the collision-aware resolver.
- [ ] Structured warn log on collision fallback and/or graph-validation soft error for duplicate StorageLayerIds.
- [ ] JS compat lane evidence attached (MapLibre/OpenLayers green against a duplicate-binding seed).

## Affected Repo(s)

honua-server

## Gate Tier Impact

PR gates (build, test); nightly JS compatibility lanes.

## Release/Deploy Impact

None.

## Non-Goals

Forbidding duplicate StorageLayerIds outright (treated as a legitimate compatibility-publication pattern).

Related to #2759, PR #2774.

### LINKED ISSUE #2802: tech-debt(telemetry): static-meter metrics classes are untestable in parallel — migrate to IMeterFactory
## Problem / Summary
Metrics classes modeled on `AlertPipelineMetrics` (#2427) — including `FeatureStreamMetrics` (`src/Honua.Server/Features/Streaming/FeatureStreamMetrics.cs`) — are static classes holding process-global counters/gauges on the shared static `HonuaTelemetry.Meter`. Observable gauges read global fields (e.g. `_activeSessions`), and multiple component instances overwrite the same global with last-write-wins semantics.

Consequences:
- Tests cannot isolate a component's metrics: any parallel test that touches the same instrument pollutes observations. `FeatureStreamMetricsTests.DisconnectSession_RecordsAdminCloseAndRefreshesActiveGauge` failed ~1 in 3 merge-train batches on 2026-07-12 for exactly this reason (deflaked tactically in #2801 with deadline polling, which masks rather than removes the shared-state race).
- In any future multi-instance scenario (two session managers in one process), the gauges silently report only the last writer's count.

## Why it matters now
Foundation-lane test failures escalate entire merge-train batches (they are never attributable to a member diff), so every flaky metrics assertion has an outsized CI cost. The tactical polling fix reduces but cannot eliminate the race.

## Proposed direction
Migrate metrics classes from the static-meter pattern to `IMeterFactory` injection (the standard .NET 8+ pattern):
- `HonuaTelemetry` exposes meters via `IMeterFactory` (DI singleton) instead of a static `Meter`.
- Metrics classes become injectable singletons owning their instruments; observable gauges read instance state (or aggregate registered instances).
- Tests construct an isolated `MeterFactory`/meter per test and filter `MeterListener` by meter instance — full parallel isolation, no global state.
- OTel wiring (`AddMeter("Honua")") continues to work: IMeterFactory meters keep the same meter name.

## Acceptance criteria
- No metrics class holds mutable process-global instrument state; instruments are created via `IMeterFactory`.
- `FeatureStreamMetricsTests` (and `AlertPipelineMetricsTests` siblings) assert against per-test isolated meters with no polling/retry loops.
- OTel export surface unchanged (instrument names, units, tags) — verified by existing telemetry tests.
- Remove the #2801 deadline-polling workaround once isolation lands.

## Affected repos
honua-server

## Gate-tier impact
Test reliability improvement; no new gates.

## Release/deploy impact
None (telemetry wiring internal; exported instruments unchanged).

## Non-goals
Changing instrument names/tags or adding new metrics.

mikemcdougall:
# Pull Request

## Issue Link
Closes #2698
Closes #2796
Closes #2798
Closes #2799
Closes #2802

## Summary
Second backlog batch above #2457 (follows the pattern of #2800, no overlap with it): five independent, well-scoped units implemented in isolated worktrees, each reviewed against its issue's acceptance criteria before being merged here. Deliberately excluded: #2793/#2794/#2795 (extend code that only exists on PR #2800's branch — they follow once it merges) and #2770 (needs the live serverless/Lambda demo to root-cause).

## Changes Made
- **#2799** — OGC Maps duplicate-binding follow-through: handler-resolved resource identity now flows into the renderer via `MapRenderRequest.ResolvedLayers` (no second first-wins `ResourcesByStorageLayerId` lookup in `VectorAwareRasterMapRenderer`); new shared `OgcMapsResourceResolver` used by both the rendering and map-tiles handlers so `…/map` and `…/map/tiles` agree; structured warn log (EventId 5919) on collision fallback; unfiltered dataset-map path enumerates storage bindings collision-aware. Regression tests use the real renderer over a raster-first duplicate-binding fixture (only `IFeatureReader` stubbed) — the fixture seed is available for the nightly MapLibre/OpenLayers compat lanes.
- **#2798** — GP mutating-tier authz hardening: `ExecuteMutatingProcess` evaluated against the REQUESTING principal at workflow run creation (per expanded step, before run persistence) and at publication creation, closing the synthesized-admin orchestrator bypass before publication-run opens to non-admin operators; `ContainsMutatingProcess` now runs unconditionally in `SubmitJobAsync` (including the custom-code branch); catalog invariant tests force an explicit tier decision (exact mutating-set snapshot + structural sink/import-are-always-Mutating guard, with the `data-management.append` analytic exception documented); authz denials now carry the actual denied operation into logs; end-to-end 403/not-403 integration test pair on the OGC Processes surface with an Execute-only role.
- **#2802** — telemetry test isolation: `FeatureStreamMetrics`, `AlertPipelineMetrics`, and `OutboxMetrics` migrated from static classes on the shared static `HonuaTelemetry.Meter` to injectable `IMeterFactory`-backed singletons with instance-owned gauge state; `AddMetrics()` registered in ServiceDefaults; tests assert against per-test isolated meters via new `TestMeterFactory` and the #2801 deadline-polling workaround is removed. OTel export surface unchanged (instrument names/units/descriptions/tags byte-identical, meter name still "Honua"). Scope note: `ControlPlaneTelemetry`, `EnhancedTelemetry`, and `PerformanceMetrics` deliberately left as-is — no multi-instance last-write-wins race, no flaky tests, and a heavy call-site ripple; they can follow if the pattern proves out.
- **#2698** — raster/surface GP parity on OGC API Processes: the `gp.raster-surface` capability descriptor flips `CanServe` false→true, making the 19 `surface.*`/`raster.*` process ids directly describable and executable at `/ogc/processes/processes/{id}` exactly as their single-step `honua-geoprocessing` plan wrappers already ran them — same shared `SubmitJobAsync` pipeline, so async job semantics, tier/RBAC/approval/admission gates all apply identically (proven by a test asserting the direct path produces the same durable `ExecutionJobRecord.Spec` as the plan-wrapped path). The flip is contained: the only other `CanServe` consumer (`PortalItemProjector`) can never resolve `gp.raster-surface`. `gp.raster-conversion` (e.g. `conversion.rasterize`) intentionally stays plan-wrap-only as a possible follow-up.
- **#2796** — WCS 2.0 Zarr GetCoverage contract gaps: `RANGESUBSET` on Zarr slices now rejected explicitly with `OperationNotSupported`/locator `RANGESUBSET` instead of silently returning the primary-variable render (multi-band composition remains a non-goal); over-extent spatial trims clamp to the intersection with the coverage extent (matching the plain `IRasterStore`/CITE path) with `InvalidSubsetting` only on empty intersection, so a client echoing the DescribeCoverage-advertised extent round-trips; the oversize-output guard now reports a truthful locator (`COVERAGEID` with a down-scale hint when no scaling operator was supplied, `SCALESIZE` when one was) and is covered by a test that actually reaches it. Both intentional divergences documented in the WCS reference doc.
- `docs/gis/data/feature-catalog.json` regenerated per unit (new `[Endpoint]` proving tests registered).

## Testing
- [x] Architecture tests pass

Per-unit targeted runs were green before merging: OGC Maps duplicate-binding endpoint tests 3/3 + tile-set regression 16/16; GP authz orchestration/catalog/403 suites green; metrics suites 125/125 + architecture 127/127; processes parity suites 29 + 6 + 47 green; WCS Zarr suites 10/10 + reader 4/4; `dotnet format` clean on all changed files.

## Gate Impact
- [ ] None — no gate impact

## Docs or Contract Impact
- [ ] OpenAPI spec changed
- [x] Documentation updated

## Release/Deploy Impact

Release notes worth flagging: raster/surface GP tools become directly callable via OGC API Processes (previously plan-wrap-only); WCS Zarr over-extent trims now return the intersection instead of a 404; Zarr `RANGESUBSET` requests that previously returned wrong data now fail loudly.

## Breaking Changes
None (the WCS RANGESUBSET rejection replaces a silent wrong-data 200; the Zarr trim change is spec-conformance).

---

## Pre-PR Checklist
- [x] Ran `scripts/ci/pre-pr-check.sh` and all checks passed — first run completed the Release build for the full solution and died only on host disk exhaustion at the final test-project copy step (environmental, known ENOSPC mode); rerun is in progress locally and every unit's targeted build/test/format runs were green pre-merge. CI is authoritative for this PR.


mikemcdougall:

mikemcdougall:
Superseded by #2826, which consolidates #2816/#2818/#2820/#2825 into a single train candidate after the batch escalation in run 29264800471 (cross-PR batch CI is evaluated once there). No content dropped — all commits are merged into batch/backlog-consolidated.

PR #2825, opened 2026-07-13
fix: epic-2803 P0 batch — config secret redaction, real DB-pressure sensing, plan-contract honesty, capability honesty, MCP CRS guards

mikemcdougall:
# Pull Request

## Issue Link
Closes #2804
Closes #2805
Closes #2806
Closes #2807
Closes #2808
Related to #2803

## Summary
The five P0 release-blocker children of epic #2803 (operator + MCP surface honesty), implemented on isolated branches, reviewed per-issue, and validated together on this merged batch. This closes the epic's "do not ship / do not demo" items: the admin secret-exfiltration path, the blind DB-pressure sensing behind the self-operating story, and the silently-wrong plan/CRS answers behind the agent-analysis story.

## Changes Made
- **#2804 (security)** — admin config discovery (`GET /api/v1/admin/configuration/discover`) redaction is now deny-by-default via a shared `ConfigurationValueRedactor`: only types that cannot embed secret material (booleans, numbers, enums, timestamps/durations, GUIDs) are exposed; all strings, URIs, and structured values return `"***"` regardless of property name (the old four-substring name denylist leaked connection strings, endpoints, and usernames raw); `null` still signals "unset". Both extraction sites (typed options and the `IConfiguration` fallback) route through it; the name heuristic survives only as the informational `IsSecret` flag.
- **#2805 (ops sensing)** — the `db-bounded-admission-pressure` findings rule and ops-health snapshot now read a real signal: `QueryConcurrencyGate` (the actual admission chokepoint) implements a new Core `IDatabaseAdmissionPressureSource` (utilization, in-flight, queued waiters, windowed acquisition timeouts), consumed by a new primary `AdmissionGateDatabasePressureSignal` with the collector-backed signal demoted to an honest fallback when no gate is registered. The prior implementation read counters no production code wrote, so the rule could never fire. A new architecture guard (`FindingsSensorWriterTests`) asserts every findings-rule metric has a production writer so this class of dead sensor cannot recur.
[...]

PR #2820, opened 2026-07-13
fix: backlog batch — GP resolution bounds, registry SRID classification, GeoServices 501 mapping, MCP render_map capability errors

mikemcdougall:
# Pull Request

## Issue Link
Closes #2770
Closes #2793
Closes #2794
Closes #2795

## Summary
Third backlog batch above #2457, following #2816: the three follow-ups deferred out of PR #2800 (they extend code that first landed there) plus the MCP render_map serverless failure. Each unit was implemented on its own branch, reviewed against its issue's acceptance criteria, and merged here; validation that was interrupted mid-run was re-executed on the merged batch.

## Changes Made
- **#2793** — GP worker resolution bounds: new `GdalOutputGridGuard.TryAdmitResolution` bounds the `-tr`-derived output grid (extent ÷ cell size, per-axis + pixel-count + Float64-byte caps, same `GdalWorkerOptions` knobs as #2782) before the subprocess spawns; `raster.resample` derives the input extent from declared pixel dimensions × GeoTIFF ModelPixelScale, `conversion.rasterize` from the untrusted GeoJSON payload envelope via a new forward-only `Utf8JsonReader`-based `GdalVectorEnvelopeReader` (no geometry allocation). Un-derivable extents admit and stay backstopped by timeout/artifact caps (documented); the residual note in the guard's remarks is removed.
- **#2794** — registry-backed geographic-SRID classification: new `IGeographicSridClassifier` + `RegistryGeographicSridClassifier` (DI singleton) answer geographic-ness from `ICrsRegistry`/`spatial_ref_sys`-derived `CrsDefinition.IsGeographic`, with the static `GeographicSridClassifier` list demoted to bootstrap/fallback (registry absent, SRID unknown, or lookup failure). Consumed by the ImageServer coordinate-metadata and measure handlers and the tile-matrix registry; `BoundingBox`/`SpatialReference` static call sites documented as fallback-tier. The registry path also resolves geocentric codes in the 4000–4999 block from WKT, superseding the range heuristic's conservative exclusion set where a registry is present.
[...]

PR #2818, opened 2026-07-13
ci: prune non-running test projects from pre-pr build set

mikemcdougall:
# Pull Request

## Issue Link
Closes #2817

## Summary
Makes `scripts/ci/pre-pr-check.sh` prune test projects that neither run in the invocation nor contain changes from its build set. Test projects are dependency leaves — nothing references them — so dropping them from the throwaway solution filter removes exactly their own compile + `bin/` materialization (~1–2 GB each: full dependency tree + native runtime assets) while every src project still compiles, either listed or pulled transitively as a `ProjectReference` of a kept test project. This is where the gate's worst case went: a `Honua.Core`/`Honua.Hosting` edit closes over effectively all 76 projects, and on 2026-07-12 two runs filled the shared dev disk to 100% (ENOSPC masquerading as MSB3021/MSB3027) building suites the script never executes. CI is unaffected — it builds every test project in its own shards, so no warnings-as-errors coverage is lost overall.

Measured build-plan effect (via the new debug flag): a provider-scoped diff drops 27 → 17 projects (SMART) and → 8 (FAST); a force-full-scope diff still prunes 15 of 76.

## Changes Made
- Shard selection (`RUN_ALL_SHARDS`/`SELECTED_SHARDS`, including the no-`csproj` → `Honua.Server.Tests` monolith fallback mirroring `run-server-test-shard.sh`) is resolved before the build step, and the build set keeps only: non-test projects, the local run set (4 unit projects, LoadTests, Architecture.Tests, selected shard projects), and test projects with changes of their own (their compile is the local signal).
- The SMART-mode `AFFECTED=ALL` case (force-full paths such as props/CI files changed) now also builds via a pruned solution filter over all src projects instead of the whole solution; explicit `HONUA_PRE_PR_FULL=1` retains the true build-everything behavior with no pruning.
- `dotnet restore` is scoped to the solution filter in SMART mode instead of always restoring `Honua.sln`.
- Format verification scopes by FULL (not by `AFFECTED`): a force-full build scope no longer forces a full-solution `dotnet format`, since `CHANGED_CS` is still authoritative in SMART mode; dropped `--verbosity diagnostic` (per-document logging across 76 projects is minutes of console I/O; failures still name each unformatted file).
- New `HONUA_PRE_PR_PRINT_BUILD_PLAN=1` debug flag prints the resolved build set and exits before restore/build.
- Header docs updated for the pruning behavior and the new flag.

## Testing
- [x] Unit tests added/updated
[...]

Issue #2732, opened 2026-07-11
tech-debt: unify geographic-SRID classification — four divergent allowlists drive axis/units/measure decisions


mikemcdougall:
### Problem Summary

At least four independent "is this SRID geographic?" classifiers coexist and disagree:

- `BoundingBox.IsGeographicSrid` (`src/Honua.Core.Abstractions/Features/Shared/Models/BoundingBox.cs:470-487`) — ~20 codes, self-described "single source of truth"; drives WMS axis order and scale math via `SpatialReference.IsGeographic`.
- `SpatialConstants.GeographicSrids` (`SpatialConstants.cs:73-82`) — 7 codes (includes 4759, which BoundingBox lacks); drives `TileMatrixSetRegistry.IsGeographicSrid` (custom gridsets), `DistanceConversions.IsGeographicSrid` (DuckDB distance branch, Postgres KNN geodesic gate).
- `ImageServerMeasureHandler` — literal `== 4326` (distance) and `is 4326 or 4269` (area) — a third and fourth list in one file.
- `PostgresGeometryOperationService.IsLikelyGeographicSrid` (`:465-466`) — 6 codes, different set again (has 4619, lacks 4617/4759); only a spatial_ref_sys-miss fallback (the primary path there correctly derives geographic-ness from `spatial_ref_sys` WKT/proj4 — the pattern the rest should adopt).

Downstream majors caused by this fragmentation (each also listed on its own issue where user-facing): DuckDB meters-as-degrees, Postgres KNN/returnDistance in degrees, ImageServer degrees-as-meters, WMS 1.3.0 axis order wrong for non-whitelisted geographic CRSs, custom gridset on EPSG:4490 classified projected → WMTS TopLeftCorner axis order diverges from the rest of the server. Related constant drift: 3+ duplicated Web Mercator alias tables alongside canonical `SpatialReferenceExtensions.NormalizeWebMercatorSrid`, and duplicate Web Mercator constants at divergent precision (`TilePackageWriter.cs:35`).

### Why This Matters Now

Every new geodesy fix keyed off "geographic vs projected" picks one of four disagreeing answers; unifying removes the root cause of 5+ major bugs at once.

### Acceptance Criteria

- One registry-backed classification service (derive from `spatial_ref_sys`/`ICrsRegistry`, static list only as bootstrap/fallback) consumed by BoundingBox/SpatialConstants/DistanceConversions/ImageServer/TileMatrixSetRegistry call sites.
[...]

PR #2826, opened 2026-07-13
fix: consolidated backlog batch — waves 1-3 (geodesy/WCS/maps/authz/telemetry, GP hardening, epic-2803 P0s, pre-pr CI)

### LINKED ISSUE #2698: Raster/surface process ids excluded from FirstSliceAutomatedVectorProcessIds — 404 on direct process-id execution

### LINKED ISSUE #2770: bug(mcp): honua_render_map returns generic internal error for all layers on the serverless/Lambda demo

### LINKED ISSUE #2793: GP worker: bound caller-controlled output resolution (-tr/cellSize) in resample and rasterize

### LINKED ISSUE #2794: tech-debt: registry-backed geographic-SRID classification (static allowlist is the only source outside Postgres)
## Problem Summary

Follow-up to #2732 (parent epic #2729). The canonical `GeographicSridClassifier` unified the four divergent allowlists, but the issue's acceptance criterion — *derive classification from `spatial_ref_sys`/`ICrsRegistry`, static list only as bootstrap/fallback* — was deliberately deferred: the classifier is a static list with no registry seam, and every non-Postgres consumer (WMS/WFS axis order via `BoundingBox.IsGeographicSrid`, ImageServer mensuration, tile/render paths) answers from the 21-code list (plus the documented 4000–4999 range heuristic for ImageServer). Genuinely geographic codes outside the list still misclassify as projected on those paths even when the registry knows better. The review also noted the geocentric-exclusion set in `IsGeographicOrUnlistedGeographicRangeSrid` (4936/4978/4984) is a conservative subset pending a complete enumeration.

## Why This Matters Now

This is the remaining root cause behind the epic-#2729 misclassification family; every new geodesy fix currently inherits the static list's blind spots.

## Acceptance Criteria

- [ ] A DI-reachable registry-backed classification service (derive geographic-ness from `ICrsRegistry`/`CrsDefinition`/`spatial_ref_sys` WKT), with the static list as bootstrap/fallback only, consumed by the DI-reachable call sites (ImageServer handlers, tile registry, providers).
- [ ] Static-only call sites (`BoundingBox`, `SpatialReference`) documented as fallback-tier, or refactored to accept an injected answer where feasible.
- [ ] Complete geocentric-code enumeration for the 4000–4999 range heuristic (or retire the heuristic once registry answers cover its consumers).
- [ ] `GeographicSridClassifier` docs updated to reflect the delivered architecture.

## Affected Repo(s)

honua-server

## Gate Tier Impact

Fast tier.

## Release/Deploy Impact

None.

## Non-Goals

Changing the narrow `GeodesicDistanceSafeSrids` provider policy (#2731 semantics).

Parent: #2729.

### LINKED ISSUE #2795: GeoServices: not-implemented operations surface as error.code=500; consolidate remaining private error-assertion helpers
## Problem Summary

Two follow-ups surfaced by the #2778 review:

1. **Product**: `GeoServicesErrorCodes.FromHttpStatusCode` has no `501` branch, so every not-implemented operation (BandArithmetic/ExtractBand exports, unsupported pixelType, samples without a Zarr store, identify variants) emits body `error.code = 500`. Clients and SDKs branching on `error.code` cannot distinguish 'operation unsupported — change parameters' from 'internal server fault'. Seven #2778 tests now pin the current 500 behavior and carry comments noting the collapse; they will need updating together with the mapping fix (Esri REST convention suggests a client-actionable 400-family error with a specific message, or a preserved 501).

2. **Test-plumbing**: ~20 GeoServices test files still declare private `AssertGeoServicesErrorAsync` copies whose signatures differ from the shared `GeoServicesErrorAssertions` helper (#2778) that is now globally imported. The private members shadow the global import; deleting a local copy later silently rebinds call sites to the shared helper's different exact-vs-BeOneOf semantics. Convert them onto the shared helper deliberately.

## Why This Matters Now

The 500-for-501 mislabel is a live client-facing contract wart on the public GeoServices surface; the helper shadowing is a latent test-semantics trap.

## Acceptance Criteria

- [ ] `FromHttpStatusCode` maps 501 to a deliberate, documented body code (decision: 501 pass-through vs 400 with 'operation not supported' message) consistent with Esri REST conventions.
- [ ] The 7 pinning tests updated alongside, with the mapping decision documented.
- [ ] Remaining private `AssertGeoServicesErrorAsync` copies removed in favor of the shared helper (per-file conversion verifying expected codes stay faithful).

## Affected Repo(s)

honua-server (possibly honua-sdk-js/-dotnet/-python release notes if SDKs branch on error.code)

## Gate Tier Impact

Fast tier.

## Release/Deploy Impact

Client-visible error-code change for not-implemented operations — release note required.

## Non-Goals

Changing the 200-with-body GeoServices error transport (spec-mandated).

### LINKED ISSUE #2796: bug(wcs): Zarr GetCoverage ignores RANGESUBSET and 404s over-extent trims instead of intersecting

### LINKED ISSUE #2798: GP authz: workflow-run surface bypasses the mutating-process tier via the synthesized admin orchestrator principal

### LINKED ISSUE #2799: bug(ogc-maps): duplicate-binding fix stops at the handler — renderer re-resolves via first-wins index; map/tiles siblings still 404; collisions stay silent

### LINKED ISSUE #2802: tech-debt(telemetry): static-meter metrics classes are untestable in parallel — migrate to IMeterFactory

mikemcdougall:
# Pull Request

## Issue Link
Closes #2698
Closes #2770
Closes #2793
Closes #2794
Closes #2795
Closes #2796
Closes #2798
Closes #2799
Closes #2802
Closes #2817
Related to #2807
Related to #2729
Related to #2803

## Summary
Consolidated backlog batch (waves 1–2 + the pre-pr CI fix), rebased over the newly-landed #2819. Scope note: the original revision of this PR also carried the epic-#2803 P0 batch (#2804–#2808, formerly PR #2825); #2819 landed a parallel implementation of those on trunk first, so that content was dropped and this PR now carries only the work that remains unique. One #2819 integration is included: trunk's new `honua.alerts.evaluation_no_leader` gauge was ported into #2802's IMeterFactory instance pattern (`AlertEvaluationBackgroundService` now takes the injected `AlertPipelineMetrics`), and #2819's `sourceCrs` parameter prose was reworded ("e.g. EPSG:3857") because it tripped the #2732 SRID-allowlist architecture ratchet — trunk is latently red on that guard today; this PR greens it.

All Codex/code-quality review findings from the previous revision are addressed: the P2 envelope-reader finding is fixed (a `properties` subtree is now skipped wholesale, so feature attributes named `coordinates`/`bbox` cannot pollute the derived extent — regression test added); integer-division precision in `GdalOutputGridGuard` fixed; two `foreach`+`if` filters converted to explicit LINQ; test float arithmetic moved to float literals. The two MemoryStream-dispose findings are false positives: those fixture streams are returned via `Task.FromResult<Stream>` and ownership transfers to the Zarr reader, which disposes them.

## Changes Made
- **#2698** — raster/surface GP process ids directly describable/executable via OGC API Processes (`gp.raster-surface` CanServe flip; direct path proven to produce the same durable `ExecutionJobRecord.Spec` as the plan-wrapped path; async job semantics and tier/RBAC gates identical).
- **#2796** — WCS Zarr GetCoverage: `RANGESUBSET` rejected explicitly (`OperationNotSupported`/locator `RANGESUBSET`) instead of silently returning the primary-variable render; over-extent spatial trims clamp to the intersection with the coverage extent (`InvalidSubsetting` only on empty intersection; advertised extent round-trips); oversize guard reports a truthful locator and is covered by a test that reaches it; divergences documented.
- **#2798** — GP mutating-tier authz: `ExecuteMutatingProcess` evaluated against the requesting principal at workflow run creation (per expanded step, pre-persistence) and publication creation; `ContainsMutatingProcess` unconditional in submit (custom-code branch included); exact + structural catalog tier-invariant tests; denial logs carry the real operation; end-to-end 403/not-403 pair.
- **#2799** — OGC Maps duplicate-binding completion: resolved resource identity flows into the renderer (`MapRenderRequest.ResolvedLayers`; no second first-wins lookup), shared collision-aware `OgcMapsResourceResolver` for the map and map/tiles handlers, collision warn log (EventId 5919), collision-aware unfiltered dataset-map enumeration; real-renderer regression fixture.
- **#2802** — `FeatureStreamMetrics`/`AlertPipelineMetrics`/`OutboxMetrics` migrated to IMeterFactory-injected singletons with instance-owned gauge state; per-test isolated meters via `TestMeterFactory`; #2801 deadline-polling removed; OTel export surface byte-identical (trunk's new `evaluation_no_leader` gauge ported into the pattern).
- **#2770** — MCP render_map: native rasterizer load failures (missing/wrong-arch libSkiaSharp etc.) detected via `RasterRenderingUnavailableException`, logged with the real exception server-side, and mapped to a `failed_precondition` capability error instead of the generic `internal` envelope; empty-render message now names the no-data case. Remaining work is the Lambda image's native render dependencies (honua-iac).
- **#2793** — GP worker bounds resolution-derived output grids pre-spawn: `GdalOutputGridGuard.TryAdmitResolution` (extent ÷ cellSize vs the existing width/height/pixel/byte caps) on the resample `-tr` path (extent from declared dims × ModelPixelScale) and the rasterize `cellSize` path (extent from a forward-only `Utf8JsonReader` GeoJSON envelope scan that skips `properties` subtrees); un-derivable extents stay backstopped by timeout/artifact caps.
- **#2794** — registry-backed geographic-SRID classification: `IGeographicSridClassifier`/`RegistryGeographicSridClassifier` answer from `ICrsRegistry` WKT-derived definitions with the static `GeographicSridClassifier` as bootstrap/fallback; consumed by ImageServer coordinate/measure handlers and the tile-matrix registry; static-only call sites documented as fallback-tier.
- **#2795** — GeoServices not-implemented operations return body `error.code = 501` (deliberate pass-through; Esri `error.code` already carries non-HTTP values) instead of collapsing to 500; the seven pinning tests updated; remaining private `AssertGeoServicesErrorAsync` copies removed in favor of the shared helper.
- **#2817** — pre-pr-check builds only what the run executes (prunes non-running, unchanged test projects; restore scoped to the solution filter; format scoped to changed files; `HONUA_PRE_PR_PRINT_BUILD_PLAN=1` debug flag).

## Testing
- [x] Architecture tests pass

On this rebased assembly: `Honua.Worker.Gdal.Tests` 271 passed (includes the new envelope-reader regression test); alerts/streaming/outbox metrics suites 72/72 (validates the gauge port); feature-catalog regeneration produces a committed, emitter-verified catalog; full architecture suite green after the prose fix (the one failure was trunk-inherited from #2819, detailed above); `dotnet format --verify-no-changes` clean on all changed files. Per-unit evidence for each issue is in the closed source PRs (#2816/#2820).

## Gate Impact
- [ ] None — no gate impact

## Docs or Contract Impact
- [ ] OpenAPI spec changed
- [x] Documentation updated

## Release/Deploy Impact

Release notes: GeoServices body `error.code = 501` for not-implemented operations (was 500 — SDKs should treat as "change the request, don't retry"); raster/surface GP tools directly callable via OGC API Processes; WCS Zarr over-extent trims return the intersection instead of 404; GP worker rejects over-cap `-tr`/cellSize resolutions pre-spawn; MCP render_map returns `failed_precondition` when the runtime image lacks native render libraries.

## Breaking Changes
None (behavior changes are safety-hardening or spec-conformance, flagged above for release notes).

---

## Pre-PR Checklist
- [x] Ran `scripts/ci/pre-pr-check.sh` and all checks passed — targeted equivalent on the rebased assembly (affected builds, affected suites, catalog emitter, full architecture suite, format), detailed above; CI authoritative.

Supersedes #2816, #2818, #2820 (closed; content carried here). #2825 was dropped — its scope landed on trunk via #2819.


mikemcdougall:
Review findings from the previous revision — disposition on the rebased branch (f6dc916):

- **GdalVectorEnvelopeReader (Codex P2)**: fixed — the reader now skips a `properties` member's entire subtree, so feature attributes named `coordinates`/`bbox` can no longer contribute to the derived extent. Regression test added (`TryReadEnvelope_AttributeNamedCoordinatesOrBbox_DoesNotWidenEnvelope`).
- **GdalOutputGridGuard integer division**: fixed — byte-cap comparison now uses floating-point division.
- **RasterRenderingUnavailableException / VectorAwareRasterMapRenderer implicit filters**: fixed — converted to explicit LINQ filters.
- **Wcs20ZarrTrimEndpointsTests float arithmetic**: fixed — float literals.
- **MemoryStream not disposed (Wcs20ZarrOversizeEndpointsTests:175, Wcs20ZarrTrimEndpointsTests:224)**: not changed — these are intentional ownership transfers: the fixture returns the stream via `Task.FromResult<Stream>` and the Zarr slice reader disposes it; disposing before return would hand the caller a dead stream.
- **PlanExecutabilityAnalyzer / ManagedBufferAggregateExecutor findings**: moot — those files left this PR when the epic-#2803 P0 content was dropped in favor of the parallel implementation that landed on trunk via #2819.

Also fixed in passing: #2819's `sourceCrs` parameter prose tripped the #2732 SRID-allowlist architecture ratchet (`SourceTree_ShouldNotReintroduceLocalSridAllowlists`) — trunk is currently red on that guard; this branch greens it with a one-word reword.

mikemcdougall:

mikemcdougall:

PR #2829, opened 2026-07-14
fix(import): resolve CRS84, legacy Esri wkid, and GML axis order

### LINKED ISSUE #2729: epic: geodesy deep-audit 2026-07 — SR/transform, projection, topology correctness (tracking)
### Problem Summary

A deep geodesy audit of trunk `26bcca060` (2026-07-11; six parallel review passes over core projection math, protocol CRS semantics, provider spatial SQL, topology/geometry operations, tiles/rendering, and geoprocessing executors, with hand verification of all critical findings) surfaced 4 critical clusters, ~20 majors, and ~30 minors. This epic tracks the child issues.

Verified-solid foundations (no action): spherical 3857/ECEF/UTM-MGRS math is formula-exact; tile-matrix constants match OGC 17-083r4 digit-for-digit in all four definitions; OGC API Features Part 2 CRS handling is compliant; `areasAndLengths` geodesic is true ellipsoidal (PostGIS geography); edit-path validity pipeline covers all four interactive protocols; no SRID injection; filter-side (index-preserving) transforms with antimeridian bbox splitting.

### Why This Matters Now

Several defects silently return wrong query results or wrong measurements (no error, HTTP 200) on specific providers/CRSs, which corrodes trust in every protocol surface built on the shared pipelines.

### Child Issues

- [ ] #2730 — MySQL `Within`/`Contains` inversion (+ canonical enum doc fix)
- [ ] #2731 — DuckDB `ST_Transform` axis order + meters-as-degrees distance
- [ ] #2732 — Unify geographic-SRID classification (tech-debt, root cause of several majors)
- [ ] #2733 — Geoprocessing catalog unit dishonesty (meters/geodesic vs planar executors)
- [ ] #2734 — ImageServer mensuration planar-as-meters
- [ ] #2736 — GeoServices query path rejects `outSR=102100`
- [ ] #2737 — WFS/FES CRS input defects (legacy bbox axis, srsName defaults, URN fallback)
- [ ] #2738 — WMTS WorldCRS84Quad TopLeftCorner axis swap
- [ ] #2739 — Antimeridian extent transform + WMS GetFeatureInfo wrapped bbox
- [ ] #2740 — Postgres distance defects (storage-mapped ::geography 500, KNN degrees, non-sargable DWithin)
- [ ] #2741 — SQL Server geography ring orientation
- [ ] #2742 — GeometryService semantics (empty→400, DE-9IM touch patterns, autoComplete)
- [ ] #2743 — Import geodesy gaps (validity gate, WKT2/ESRI .prj/GPKG srs_id CRS detection)
- [ ] #2744 — Z/M loss (reproject executors, shapefile export)
- [ ] #2745 — Ring winding on secondary writers (GeoJSON/GML/KML)

### Minor findings not individually ticketed

Silent unit fallback to meters in GeometryService request parser; 39.37 vs 39.3701 in/m seam between advertised LODs and scale-visibility check (layer hidden at exactly its advertised LOD); custom-gridset validator doesn't check cellSize↔scaleDenominator consistency; WMTS TopLeftCorner formatted F6 (custom-gridset precision); gdalwarp silently drops invalid `sourceSrs` (target fails loudly); PDAL treats EPSG:4269 as WGS84 passthrough (datum conflation ~1-2 m); EWKB-embedded SRID silently overwritten by declared srid in GP executors; `calculationType=preserveShape` silently computed as geodesic; trimExtend `extendHow` parsed but never consumed; offset bevelRatio clamped to NTS mitre default; empty geometries keep source SRID through reproject; STAC extent degrades to world bbox on transform failure; OData legacy `crs` member emission for non-4326 layers; `1<<level` overflow at level≥31 (unreachable, z capped at 24); WMS 1.1.1 exception code `InvalidCRS` instead of `InvalidSRS`; three+ duplicated Web Mercator alias tables; ~8 protocol-local CRS identifier parsers with divergent fallback behavior.

### Acceptance Criteria

- All child issues resolved or explicitly deferred with rationale.
- Cross-provider regression tests lock the `SpatialRelationship` enum semantics for every provider.

### Affected Repo(s)

honua-server

### Gate Tier Impact

Fast + Slow integration tiers (new provider/protocol regression tests expected).

### Release/Deploy Impact

No deploy/config changes; correctness fixes only.

### Non-Goals

New projection engine / PROJ integration; geodesic implementations beyond what the catalog already advertises.

mikemcdougall:
Closes #2729

## Summary

Three file-import CRS-detection correctness fixes surfaced by the Geodesy Deep-Audit Epic (#2729), covering the "unreliable CRS detection" and axis-order clusters.

### 1. GML `CRS84` recognized as EPSG:4326
`ParseSrsNameToSrid` now resolves every OGC CRS84 spelling (`CRS84`, `CRS:84`, `urn:ogc:def:crs:OGC:1.3:CRS84`, `http://www.opengis.net/def/crs/OGC/1.3/CRS84`) to 4326. The check runs **before** the generic trailing-integer parse, which also fixes a latent bug where `CRS:84` mis-parsed to SRID **84**. Previously a conformant CRS84 GML failed with "Source SRID is required."

### 2. Esri JSON legacy `wkid` remapped
ArcGIS commonly emits a bare legacy `wkid` (no `latestWkid`) with the Web Mercator aliases `102100` / `102113` / `900913`, none of which exist in PostGIS `spatial_ref_sys` by default — so the import was rejected as an unregistered SRID. These now map to EPSG:3857. Registered EPSG codes pass through unchanged and `latestWkid` still wins over `wkid`.

### 3. GML CRS-dependent axis order (the audit's finding 1)
The GML reader previously read ordinates as-written and never performed the authority axis swap, so a GeoServer-style document declaring `srsName="urn:ogc:def:crs:EPSG::4326"` (authority order = lat/lon) imported **transposed**.

Coordinates are now swapped from authority (lat/lon) order into the internal easting/northing (X=lon, Y=lat) storage order when — and only when — the `srsName` uses an axis-order-carrying authority form (a `urn:…` URN or a `…/def/crs/…` OGC URI) **and** the CRS is geographic per the canonical `GeographicSridClassifier`. Unswapped: the legacy short form `EPSG:4326`, the `…/gml/srs/epsg.xml#4326` URL, all `CRS84` spellings, and every projected CRS. This matches GDAL's `SWAP_COORDINATES=AUTO` and GeoServer, and preserves the common ogr/QGIS lon/lat default that already worked.

Implementation: a `GmlSrs(int? Srid, bool SwapAxes)` context is threaded through the parsers (replacing the bare `int? collectionSrid`); the swap is applied once at the leaf coordinate parsers (Point/LineString/LinearRing) so nested Polygon/Multi*/GeometryCollection geometries are never double-swapped.

**Concrete result:** `urn:ogc:def:crs:EPSG::4326` with `<gml:pos>37.5 -122.1</gml:pos>` (lat lon) now imports as `POINT(-122.1 37.5)` instead of the transposed `POINT(37.5 -122.1)`.

## Correctness notes
- The x/y swap is a reflection that flips ring orientation, but import repairs invalid geometry and all emitters normalize winding (`ST_ForcePolygonCCW`); Streaming re-stamps only the SRID, not coordinates — no corruption.
- Conservative limitation (not a regression): `GeographicSridClassifier` is a fixed ~20-entry list, so a URN-declared geographic CRS outside it degrades to the prior pass-through rather than swapping — consistent with how the rest of the codebase decides axis order.

## Tests
- New: GML axis-order cases (URN geographic swaps point + polygon ring exactly once; short-form EPSG, URN projected, and CRS84 do **not** swap), Esri legacy-wkid theory, CRS84 `ParseSrsNameToSrid` cases.
- `Honua.Geometry` builds clean (0/0). All 33 GML + EsriJson reader tests pass; full `Features.Import` suite green (511 passed, 2 skipped).

🤖 Generated with [Claude Code](https://claude.com/claude-code)


mikemcdougall:

PR #2833, opened 2026-07-14
feat(wcs): advertise and validate transformable output CRS values

mikemcdougall:
# Pull Request

## Issue Link
Closes #2664

## Summary
WCS 2.0 already reprojects `GetCoverage` output (`OUTPUTCRS`/`SUBSETTINGCRS`/`BBOXCRS` → `RasterQuery.OutputSrid` → `ST_Transform`), but it never advertised which output CRS values were transformable and validated them only syntactically — so a parseable-but-unknown code such as `OUTPUTCRS=EPSG:99999` reached PostGIS and surfaced as an HTTP 500 instead of a protocol-correct WCS 400.

This PR closes both gaps by building a single bounded set of transformable CRS values (each visible coverage's native CRS plus the default WGS84/WebMercator identifiers, filtered to what the `ICrsRegistry` can resolve) and using it as one source of truth for both advertisement and validation. It mirrors the existing OGC API Coverages pattern (`OgcCoveragesHandler`).

The change is deliberately additive for CITE: the CRS values are emitted inside the `wcs:ServiceMetadata` `wcs:Extension` slot (an `xs:any` slot in `wcsAll.xsd`), and the WCS CRS-extension conformance class (OGC 11-053r1) is **not** declared in `ows:Profile`/OperationsMetadata. WCS core stays 82/82.

## Changes Made
- Injected `ICrsRegistry` into `Wcs20Handler` and added `ResolveSupportedCrsAsync`, which seeds `{native} ∪ {EPSG:4326, EPSG:3857}`, keeps only registry-resolvable SRIDs, and returns both the SRID set (validation) and the ordered EPSG-URI list (advertisement).
- `GetCapabilities`: emit `crs:crsSupported` values inside a `wcs:ServiceMetadata`/`wcs:Extension` child; declared the `crs` namespace on the capabilities root. Additive only — no conformance class advertised.
- `GetCoverage`: threaded the resolved supported-SRID set into `TryResolveRequestCrs`; a well-formed but non-transformable `OUTPUTCRS`/`SUBSETTINGCRS`/`BBOXCRS` now returns 400 `OutputCrs-NotSupported`/`SubsettingCrs-NotSupported` (OGC 11-053r1) with the correct locator; malformed values still return `InvalidParameterValue`; the native CRS default is always accepted.
- Added `CrsNamespace`, `DefaultCrsIdentifiers`, and the two exception-code constants to `Wcs20Utilities`.
[...]

PR #2862, opened 2026-07-15
feat(mcp): serve RFC 9728 protected-resource metadata for /mcp

mikemcdougall:
# Pull Request

## Issue Link
Closes #2849

## Summary

Serves OAuth 2.0 Protected Resource Metadata (RFC 9728) for the `/mcp` resource, so a spec-compliant MCP client can discover which authorization server guards the surface instead of being configured out-of-band.

The document is a projection of the OIDC authority configuration that already drives authentication (`Oidc:*` → `OidcAuthenticationOptions` → `OidcProviderCatalog`), not a second parallel config surface: the advertised `authorization_servers` are exactly the issuers the host already trusts, so the two cannot drift.

Per capability-honesty (#2803) the endpoint is **absent, not empty**, when no authorization server is configured — the route is never mapped, so routing answers `404`. An empty `authorization_servers` array would assert "this resource is guarded by no authorization server", which is a different and false claim. For the same reason the document omits `scopes_supported` and `bearer_methods_supported`: the surface does not accept bearer tokens yet (#2850), so advertising a token vocabulary the runtime does not enforce would be exactly the advertised-vs-actual gap #2803 exists to close.

Scope note on the `WWW-Authenticate` criterion: `/mcp` is deliberately anonymous-reachable and today surfaces authentication failures as JSON-RPC `unauthenticated` errors — there is currently no code path that returns HTTP 401. The challenge is therefore stamped by a response hook armed on every MCP transport route: whenever the surface returns a 401 it carries `Bearer resource_metadata="…"`. That satisfies the criterion for every 401 the surface can emit without implementing token acceptance, which is #2850's job.

## Changes Made
- Added `McpProtectedResourceMetadata` — derives `authorization_servers` from `OidcAuthenticationOptions` via the existing `OidcProviderCatalog`, resolves the `/mcp` resource identifier through the shared `BaseUrlResolver`, and constructs the well-known URL per RFC 9728 §3 (path inserted **between host and path**: `/.well-known/oauth-protected-resource/mcp`, not appended to `/mcp`).
- Added `McpProtectedResourceMetadataEndpointExtensions` — maps `GET /.well-known/oauth-protected-resource/mcp` only when a valid authority is configured, and stamps the RFC 9728 §5.1 `WWW-Authenticate: Bearer resource_metadata="…"` challenge on any 401 from `/mcp`.
- Wired both into `MapMcpOperatorSurface`; the challenge is applied to the POST/GET/DELETE `/mcp` routes via an endpoint filter.
- Registered the route in `EndpointRegistry.CatalogExtras.cs`, added the surface to the `mcp-operator` public-interface proof ledger, and regenerated `docs/gis/data/feature-catalog.json` in Release.

> **Note on the catalog diff:** regenerating `feature-catalog.json` emits the whole artifact, so the diff also contains WCS proving-test entries that were already stale on `trunk` from #2833/#2834 (the pre-existing `FeatureCatalogDriftTests` failure). Those lines are not from this change — they are what a correct regeneration produces. If the separate fix for that drift lands first, this branch will be rebased and the catalog regenerated on top of it.

## Testing
- [x] Architecture tests pass

`tests/dotnet/Honua.Ai.Tests/Source/McpProtectedResourceMetadataTests.cs` (9 tests, all passing in Release):
- **Metadata shape vs the RFC** — `200`, `application/json` (§3.2), required `resource` member (§2) equal to the `/mcp` resource identifier, `authorization_servers` carrying the configured issuer, `resource_name` present, and `scopes_supported`/`bearer_methods_supported` asserted **absent**.
- **Derivation** — with Azure AD + generic OIDC configured, `authorization_servers` equals both derived authorities, proving the document mirrors the OIDC config rather than a parallel surface.
- **Absent when unconfigured** — `404` with no authorization server.
- **Well-known URI construction** — insertion between host and path, and terminating-slash removal for a root resource.
- **`WWW-Authenticate`** — the challenge carries a quoted `resource_metadata` parameter pointing at the metadata URL, and no challenge is emitted when nothing is configured.

## Gate Impact
- [ ] None — no gate impact

## Docs or Contract Impact
- [ ] OpenAPI spec changed
- [x] Documentation updated

`docs/gis/data/feature-catalog.json` (generated) and `docs/gis/data/public-interface-proof.json`.

## Release/Deploy Impact
- [x] Requires environment variable or secret changes
- [ ] None — standard merge-and-release flow

No new configuration is introduced. The endpoint appears only where the existing `Oidc:*` authority configuration is already set; hosts with `Public:BaseUrl` configured get a stable public resource identifier.

## Breaking Changes
None. The endpoint is additive and absent unless OIDC is configured; the `WWW-Authenticate` header only appears on 401 responses, which `/mcp` does not currently produce.

---

## Pre-PR Checklist


mikemcdougall:

mikemcdougall:
**Batch CI caught a real defect here — this PR is the reason the train escalated the whole batch.** Keeping `train:escalated` on this one; the other four were innocent and have been cleared.

Batch run [29460825726](https://github.com/honua-io/honua-server/actions/runs/29460825726), shard `Server Tests (STAC and API Governance)` — exactly one failing test, exactly one stale entry:

```
Honua.Server.Tests.Features.API.EndpointRegistryDriftTests.AllDeployedEndpoints_AreTrackedInRegistry
  Expected collection to be empty because every HTTP endpoint in EndpointRegistry.All
  must correspond to a deployed endpoint.
  Stale entries: GET /.well-known/oauth-protected-resource/mcp
```

**This is my fault, not a mistake in the implementation.** I set the acceptance criterion *"the endpoint is absent (not empty) when no authorization server is configured"* (capability honesty, #2803). This PR implements that correctly — the route is only mapped when a valid OIDC authority exists. But `EndpointRegistry.All` lists it **unconditionally**, and the test environment configures no authority, so the registry advertises a route that is not deployed. The AC and the drift test are in direct conflict; the PR is caught in the middle.

Note this could not have been caught locally the way the work was verified: `EndpointRegistryDriftTests` lives in the **Server Tests** shards, not `Honua.Architecture.Tests` — so 132/132 architecture green was true and still missed it. Batch CI is the only place this shows up. Working as intended.

**The fix needs a decision, not a patch.** Options, roughly in order of preference:

1. **Make the registry entry conditional the same way the route is.** Keeps capability-honesty intact and keeps registry ≡ deployed surface. Best if `EndpointRegistry` can express conditionality — check whether other feature-gated routes already do (the run log shows `Feature 'Observability' is disabled`, so gated surfaces exist; find out how they satisfy this test).
2. **Configure an OIDC authority in the test host** so the route deploys and the drift test sees it. Also proves the endpoint actually appears when configured — arguably better coverage — but only valid if a test-host authority does not weaken the "absent when unconfigured" guarantee elsewhere.
3. **Teach the drift test about conditionally-deployed routes.** Last resort: it weakens a gate that is currently exact, and this is a repo where an ungated claim is the recurring failure mode.

**Do not** drop the "absent when unconfigured" behaviour to make the test pass — that is the whole point of the ticket, and advertising an OAuth metadata endpoint that resolves to nothing is precisely the #2803 failure mode.

Everything else in this PR stands — the RFC 9728 URI construction, derivation from the existing OIDC catalog, the `WWW-Authenticate` challenge, and the proof-ledger entry are all unaffected.

PR #2866, opened 2026-07-15
feat(wms): implement GetLegendGraphic (#2855)

mikemcdougall:
# Pull Request

## Issue Link
Closes #2855

## Summary

Implements WMS `GetLegendGraphic` on both WMS routes. It is not in the WMS 1.3.0 core spec, but every WMS client that draws a legend calls it, so a user migrating off GeoServer currently watches their legends vanish.

The central design requirement was that legends must not become a second styling path that can drift from what `GetMap` draws. This implementation guarantees that structurally rather than by convention:

- **Same style source.** The legend fetches symbology through `GetRasterStylePlanAsync` — the exact call, cache, and parse `GetMap` uses. A legend cannot show a style revision the map is not drawing.
- **Same paint resolution.** `LegendClassifier` never evaluates paint values. It enumerates the *domain* of a layer's classifying expression and emits the synthetic feature attributes that select each branch. Those attributes are fed back through the existing `StyleTranslator`/`SkiaMapRenderer` path, so a swatch is by construction the paint `GetMap` would apply to a feature carrying those attributes. Unit tests assert this equality directly.
- **No forked renderer.** `SkiaMapRenderer.RenderLegendSwatch` gained an optional `properties` argument (it previously hardcoded empty attributes, which silently collapsed every data-driven style to its fallback) and a canvas-level `DrawLegendSwatch` so multi-entry legends compose without PNG round-trips. The Esri MapServer legend path is unchanged and now shares that primitive.

Reuse of the Esri-side legend logic: the shapes align at the *swatch* level, so the shared primitive was lifted into `Honua.Hosting/Features/Rendering` where both protocol assemblies already consume the renderer — no protocol-to-protocol dependency. They diverge above that line, and correctly: Esri returns JSON with per-entry base64 swatches; WMS returns one composed raster.

Data-driven expressions resolve to discrete entries where the expression permits, and are reported rather than faked where it does not:

| Expression | Behavior |
| --- | --- |
| `match` | One entry per label (grouped labels expand), plus `Other` for the fallback arm. |
| `step` | One entry per band (`< first`, `lo - hi`, `>= last`). |
| `interpolate` over colors | One entry per ramp stop, labelled with the stop value. See note below. |
| `case` / non-attribute input | **Not representable** — no finite attribute domain to enumerate. |

`LegendURL` is advertised per-layer, not blanket-asserted (capability honesty, #2803): only layers whose resolved style plan yields painted (`fill`/`line`/`circle`) symbology are advertised. A style made only of `symbol`/`background` layers is not, because `GetMap` renders nothing for it. An integration test parses the advertised URL out of capabilities and fetches it, so the claim is proven rather than asserted.

### `interpolate` over colors is now representable (revised after #2870)

An earlier revision of this PR reported `interpolate` color ramps as unrepresentable, because `ExpressionEvaluator.InterpolateValues` coerced both endpoints through `ConvertToFloat` and resolved every ramp to **black** — sampling it would have reproduced that black (useless) or bypassed the evaluator and invented colors `GetMap` never drew (wrong).

**#2870 fixed that evaluator, so the reason no longer holds and the notice would now be stale.** The ramp is sampled at its own stops:

- A stop is the one input an `interpolate` resolves **exactly** — the evaluator returns that stop's own output verbatim rather than blending toward a neighbour. So `linear`, `exponential` and `cubic-bezier` ramps all sample identically, and the legend never has to model the curve between stops or duplicate MapLibre's interpolation math.
- Each entry therefore carries the same guarantee `match`/`step` entries carry: the swatch is the color `GetMap` paints for a feature holding that attribute value, resolved through the same `StyleTranslator` path. Two unit tests assert exactly that, including an interior stop on an `exponential` ramp (the case that would expose a sampler modelling the curve itself).

A discrete legend of a continuous ramp is a sampling, and this is the honest form of it: exact at the labelled stops, continuous between them. No `Warning` header — the entries are not approximations.

### SCALE → zoom derives through the shared `RenderZoom` (revised after #2872)

An earlier revision converted `SCALE` with a local constant, `log2(559082264.0287178 / scale)`. That constant is the **256px** GoogleMapsCompatible zoom-0 scale denominator, but MapLibre's world is `512 * 2^zoom` pixels (`Transform.tileSize = 512`), so its zoom 0 is `279541132.0143588` — exactly half. The legend was classifying at **MapLibre zoom + 1**, one level deeper than the map draws, and would have listed layers `GetMap` renders nothing for: precisely the drift this PR's architecture prevents everywhere else, leaking back in through a constant.

#2872's `RenderZoom` is now the single derivation. `SCALE` and `RenderZoom`'s bbox+pixels input are genuinely different starting points, so they are reconciled rather than forced together: a scale denominator is defined against the OGC standardized 0.28 mm pixel (`SpatialConstants.PixelSizeMeters`), which fixes a **ground resolution** and nothing else — no centre, no image size. `RenderZoom.FromScaleDenominator` rebuilds a reference extent at that resolution and defers to `FromWebMercatorExtent`; the reference pixel span cancels out of the derivation, so it imposes no camera. The legend and the map now gate on the same zoom **by construction**, with no second scale→zoom constant to drift.

The reconciliation is cross-checked against the repo's own WebMercatorQuad definition rather than against its own algebra: `TileMatrixSetRegistry` builds a level's scale denominator as `cellSize / 0.00028`, and `RenderZoomTests` already pins that a 256px tile envelope at matrix level *z* is MapLibre zoom *z-1*. `FromScaleDenominator` of that level's denominator must land on the same zoom, for levels 0/2/9/14/20 — it does. The old constant is included as a case (`559082264.0287178` → zoom **-1**, not 0), so the off-by-one cannot come back silently.

This also updates a claim that has since gone stale. An earlier revision noted that WMS `GetMap` never passed zoom to the renderer and so ignored `minzoom`/`maxzoom` (#2868). That has since been fixed on trunk — `GetMap` now derives a zoom from `BBOX` + `WIDTH`/`HEIGHT` + `CRS` through `DeriveRenderZoomAsync`. So the legend is no longer honoring a gate the map ignores: both now resolve a `RenderZoom` and gate identically, from the two inputs their respective protocols actually carry. With `SCALE` omitted the legend passes `RenderZoom.NotDerivable`, so `minzoom`/`maxzoom` do not apply and every style layer contributes.

## Changes Made
- Add `GetLegendGraphic` dispatch to `WmsRequestHandlers` and a `WmsRequestHandlers.GetLegendGraphic.cs` handler supporting `LAYER`, `STYLE`, `FORMAT` (PNG), `WIDTH`, `HEIGHT`, `SCALE`.
- Add `LegendClassifier` — projects a MapLibre style layer onto discrete legend classes (`match`/`step`, and `interpolate` sampled at its stops) or an explicit unrepresentable reason (`case`, non-attribute input).
- Add `RenderZoom.FromScaleDenominator` — reconciles an OGC scale denominator onto #2872's shared zoom derivation via the standardized 0.28 mm pixel, so scale-driven callers gate on the same zoom bbox-driven callers do. Replaces this PR's earlier local constant, which was the 256px zoom-0 denominator and gated one zoom level deep.
- Add `LegendImageComposer` — stacks swatch/label rows into one PNG; labels degrade to swatches-only where no typeface is available (the Lambda fontconfig surface from #1728).
- Thread feature attributes through `SkiaMapRenderer.RenderLegendSwatch` and extract `DrawLegendSwatch` for canvas composition.
- Advertise `LegendURL` in `GetCapabilities` only for legend-capable layers; `width`/`height` are omitted deliberately since composed size depends on label metrics.
- Register `WMS-1.3.0::GetLegendGraphic` / `WMS-1.1.1::GetLegendGraphic` in `OperationRegistry` and the public-interface proof ledger; regenerate the feature catalog in Release.
- Document the operation, the SCALE→zoom derivation, and every unrepresentable case in `docs/reference/protocols/wms-wfs-wcs-wmts.md`.

## Testing
- [x] Architecture tests pass

10 new `LegendClassifierTests` (including tests asserting classified attributes resolve to the same colors `StyleTranslator` gives `GetMap`) and 12 new `WmsGetLegendGraphicTests` covering both routes, both WMS versions, parameter handling, spec-correct `ServiceExceptionReport` errors, and the capabilities `LegendURL` round-trip. Full WMS regression suite and all 132 architecture tests pass in Release.

## Gate Impact
- [ ] None — no gate impact

`LegendURL` is added inside the existing `<Style>` element, which is core WMS 1.3.0 schema. No `<GetLegendGraphic>` element was added to `<Request>` — that requires the SLD extension namespace and would risk the 199/199 WMS 1.3 CITE pass.

## Docs or Contract Impact
- [ ] OpenAPI spec changed
- [x] Documentation updated

## Release/Deploy Impact

## Breaking Changes
None. `GetCapabilities` gains a `LegendURL` element inside `<Style>` for legend-capable layers; the element is optional in the WMS schema and additive for existing clients.

---

## Pre-PR Checklist


mikemcdougall:
**Off-by-one in the SCALE→zoom derivation — please fix before merging.** Surfaced by #2872 (WMS zoom gating, #2868), which had to derive the same quantity and verified it against MapLibre's source.

```csharp
private const double LegendScaleDenominatorZoom0 = 559082264.0287178;
zoom = Math.Log2(LegendScaleDenominatorZoom0 / scale);
```

`559082264.0287178` is the **256px** GoogleMapsCompatible zoom-0 scale denominator. **MapLibre uses 512px tiles** (`Transform.tileSize = 512`, `transform_helper.ts`), so its zoom 0 is `279541132.0143588` — exactly half.

The derivation therefore returns **MapLibre zoom + 1**. Concretely: a `SCALE` of `279541132` (MapLibre z0) yields `log2(559082264 / 279541132) = log2(2) = 1`.

**Why it matters:** the legend classifies layers at a zoom one level deeper than the map draws at. A layer with `minzoom: 10` would be listed in the legend at a SCALE where `GetMap` renders nothing — the legend advertising symbology the map doesn't draw. That is the exact drift this PR's design is otherwise structured to make impossible, leaking back in through a constant rather than through the style path.

Once #2872 lands, this also needs a one-line signature update: `ShouldRenderLayer` no longer takes `double? zoom` — it takes a required `RenderZoom` (`At(level)` / `NotDerivable(reason)`), specifically so a caller cannot silently skip gating. **The compiler will catch the call site**, which is that trap-closing working as intended. Prefer `RenderZoom`'s derivation over a local constant so the legend and the map cannot disagree by construction.

Nothing else here is affected — the classifier, the GetMap-drift seams, and the `LegendURL` gating all stand. This is one constant and one call signature.

Suggested order: land #2872 first, then rebase this and adopt `RenderZoom`. `train:hold` is on both, so neither will land unreviewed.

PR #2869, opened 2026-07-15
feat(raster): decode LZW and ZSTD COG tiles

mikemcdougall:
# Pull Request

## Issue Link
Closes #2854

## Summary

`TileDecompressor` decoded only JPEG passthrough, DEFLATE, and NONE; LZW and ZSTD were explicitly deferred. LZW is GDAL's long-standing default and dominates published imagery, so a user pointing Honua at a COG they already own had a good chance of hitting an unsupported codec in their first few minutes — on their own data, with no workaround short of re-encoding their archive.

This adds LZW (with the horizontal differencing predictor, TIFF tag 317 = 2) and ZSTD to the COG tile read path, and replaces the generic `NotSupportedException` with structured errors that name the offending codec.

The predictor is the load-bearing half of the LZW work. It is near-universal in LZW GeoTIFFs and is a *separate* failure mode from LZW itself — decoding LZW without reversing it yields plausible-looking garbage rather than a clean failure. It is threaded from the TIFF tag through `CogMetadata` to the tile resolver, so the serving path actually reverses it.

**Validated against real GDAL output, not our own encoder.** Every fixture in `tests/dotnet/Honua.Core.Tests/Raster/CogParser/Fixtures/` was produced by **GDAL 3.12.1** (via rasterio 1.5.0) with `COMPRESS=LZW` (predictor 1 and 2), `COMPRESS=ZSTD`, `COMPRESS=DEFLATE`, and uncompressed. Each `.tif` ships with a `.bin` holding **GDAL's own decode** of that file in TIFF tile order. The tests parse the real TIFF through `CogMetadataExtractor`, decode each tile, and assert byte-for-byte equality with GDAL's pixels. `scripts/raster/generate-cog-fixtures.py` regenerates them.

## Changes Made
- Added `LzwDecoder` — TIFF-flavoured LZW (TIFF 6.0 §13): MSB-first packing and the Aldus "early change" code-width steps at table entries 511/1023/2047, not the 512/1024/2048 of GIF/textbook LZW. Semantics taken from the spec and libtiff's `tif_lzw.c` rather than from memory. Uses a pooled prefix/suffix trie, so code expansion allocates no per-code strings.
- Added `TilePredictor` — reverses horizontal differencing for 8/16/32/64-bit samples, striding by `SamplesPerPixel` and honouring the file's byte order for multi-byte samples.
- Added `TilePixelLayout` — carries the tile geometry the predictor needs. `TileDecompressor.Decompress` gains an overload taking it; the existing overload is preserved and delegates with `TilePixelLayout.None`.
- Added ZSTD decode via `ZstdSharp.Port`. The frame's declared content size is used to size the output exactly, but only after bounds-checking it against the decompression-bomb ceiling — it is attacker-controlled file content. Frames without a content size (streaming-produced) fall back to the over-estimated header size and are trimmed to the bytes actually written.
- Added `UnsupportedTileCodecException` (carries `Codec` + `SupportedCodecs`) and `UnsupportedTilePredictorException` (carries `Predictor`), replacing the generic decode failure.
- Threaded predictor / bits-per-sample / endianness through `CogMetadataExtractor` (new tag 317 parse) → `CogMetadata` → `CogTileResolver`.
- Extended `TileDecompressorBenchmarks` with LZW (a real GDAL payload), ZSTD, and NONE; the DEFLATE and JPEG baselines are unchanged.

### On the new dependency

The ticket asks not to add a third-party dependency if one is already present. **`ZstdSharp.Port` is already in the solution's dependency graph** — `Parquet.Net 5.6.0` pulls it in transitively via `Honua.Geometry`. This PR promotes it to a direct reference on `Honua.Core` and pins it at **0.8.7, the version that resolves today**, so central package management does not move Parquet's zstd underneath it. (An initial 0.8.8 pin did silently bump Parquet's resolved version; that was backed out.)

It is a **fully-managed C# port of zstd** (MIT). Verified from the package contents: it ships only `lib/*/ZstdSharp.dll` with **no `runtimes/` native folder** — no native `libzstd`, and no reflection. LZW is hand-rolled, so it adds no dependency at all.

### AOT verification — what was actually run

`Honua.Server` sets `PublishAot=true`, so this is a hard constraint. Honest accounting:

- **A full NativeAOT publish could not be run on this Windows box** — ILC needs the MSVC C++ toolchain (`Platform linker not found`), and NativeAOT cannot cross-compile to the `linux-x64` RID that CI publishes. ILC never executed, so I am **not** claiming a green AOT publish.
- **What I did run:** a self-contained `win-x64` publish with `PublishTrimmed=true`, `EnableTrimAnalyzer=true`, `EnableAotAnalyzer=true`, `TrimmerSingleWarn=false`. ILLink analysed the whole closure and produced **zero IL2xxx/IL3xxx warnings mentioning ZstdSharp**. I verified this is not vacuous: `ZstdSharp.dll` survives trimming in the publish output (198 KB) with `Decompressor.Unwrap` (4 overloads) and `GetDecompressedSize` (3 overloads) retained, so it was genuinely analysed and kept rather than trimmed away unexamined.
- The pre-existing IL2xxx warnings in that run are from unrelated libraries (`System.Configuration`, etc.), not this change.

The CI AOT lane is the authoritative check and will run on the merge train.

### Non-goals honoured
No GDAL in the serving path (`src/Honua.Worker.Gdal/` untouched). No other codecs (LERC, WEBP, JPEG2000, PackBits) — they now fail with a codec-naming error. Read path only, no encoding. Predictor 3 (floating-point) did **not** fall out cheaply — it is a byte-plane transposition, structurally unlike predictor 2 — so it is explicitly rejected with a structured error rather than silently mis-decoded.

## Testing
- [x] Architecture tests pass

54 tests pass in Release (`--filter FullyQualifiedName~Raster.CogParser`). 10 GDAL fixtures cross-checked byte-for-byte against GDAL's own decode:

| Fixture | Codec | Predictor | Bands | Depth |
|---|---|---|---|---|
| `lzw_pred1_uint8` / `lzw_pred2_uint8` | LZW | 1 / 2 | 1 | 8 |
| `lzw_pred1_uint16` / `lzw_pred2_uint16` | LZW | 1 / 2 | 1 | 16 |
| `lzw_pred2_rgb_uint8` | LZW | 2 | 3 | 8 |
| `lzw_pred2_uint8_multitile` | LZW | 2 | 1 | 8 |
| `zstd_pred1_uint8` | ZSTD | 1 | 1 | 8 |
| `zstd_pred2_uint16` | ZSTD | 2 | 1 | 16 |
| `deflate_pred1_uint8` / `none_uint8` | DEFLATE / NONE | 1 | 1 | 8 |

The RGB fixture covers the multi-sample predictor stride (a distinct failure mode from single-band); the multi-tile fixture covers per-tile predictor reset and tile-offset resolution. The 16 KB fixture payloads carry enough entropy that LZW fills past 511/1023/2047 codes and hits table-full plus `ClearCode`, so every code-width transition is exercised rather than only 9-bit codes.

Guarding the guard: `Decompress_LzwWithPredictorButLayoutOmittingIt_ProducesWrongPixels` asserts the pred=2 fixture decodes **incorrectly** when the predictor is skipped, proving the GDAL comparison is load-bearing rather than vacuously green. `Decompress_SameRasterAcrossCodecs_ReturnsIdenticalPixels` cross-checks LZW/ZSTD/DEFLATE/NONE against each other. `TilePredictorTests` covers big-endian and 32-bit sample paths that GDAL (little-endian only) cannot emit.

### Benchmarks — no regression on DEFLATE/NONE

`TileDecompressorBenchmarks`, in-process, Release. This box runs many agents concurrently, and **run-to-run variance on identical code dominates any effect**:

| Path | trunk | this branch | Allocated |
|---|---|---|---|
| DEFLATE 256×256 | **33.72 µs**, then **45.75 µs** (same config, same code) | 40.26 µs | 64.33 KB both |
| NONE 256×256 | 5.86 ns | 6.75 ns | 0 B both |
| JPEG passthrough | 3.42 ns | 2.62 ns | 0 B both |

Two baseline runs of *identical* trunk code measured 33.72 µs and 45.75 µs (±36%); the branch's 40.26 µs sits **between** the two baseline samples, i.e. inside the noise band. The reliable signals are that **allocations are byte-identical** (64.33 KB) and that `DecompressZlib` is textually unchanged — the only added work on the DEFLATE/NONE paths is one inlined `HasPredictor` bool check. New paths for reference: LZW 128×128 ≈ 389 µs / 16 KB allocated, ZSTD 256×256 ≈ 22 µs.

## Gate Impact
- [ ] None — no gate impact

## Docs or Contract Impact
- [ ] OpenAPI spec changed
- [ ] Documentation updated

No routes added or changed — this is internal decode logic, so no `EndpointRegistry` drift-mirror entry, no proof-ledger entry, and no feature-catalog regeneration are required.

## Release/Deploy Impact

## Breaking Changes
None. `CogMetadata` gains three trailing optional parameters (`BitsPerSample`, `Predictor`, `IsLittleEndian`), which is source-compatible; it is not JSON-serialised and its persistence uses explicit columns. `TileDecompressor.Decompress`'s existing overload is unchanged.

> Note for reviewers: the DB metadata tier does not persist the new predictor fields. That tier is currently inert for decoding — it reconstructs overview levels with empty `TileOffsets`, which the resolver detects and falls through to a cloud scan, so tile decode always uses freshly-scanned metadata that carries the predictor. If that tier is ever activated (the PHASE-1 `ifd_cache` work), it will need the predictor persisted alongside the tile offsets or it will silently mis-decode predicted tiles.

---

## Pre-PR Checklist


mikemcdougall:

PR #2870, opened 2026-07-15
fix(rendering): interpolate colors instead of coercing stops to 0f (#2867)

mikemcdougall:
# Pull Request

## Issue Link
Closes #2867

## Summary

`interpolate` over colors rendered black. Color stops are strings (`"#f7fbff"`, `"rgba(255,0,0,1)"`), but `InterpolateValues` ran every stop through `ConvertToFloat`, which cannot parse them — so both endpoints collapsed to the `0f` default and the function returned a *number* where the renderer expected a *color*. The renderer parsed that number back as a color, failed, and fell back to black. Silently: no throw, no warning, no log. `["interpolate", ["linear"], ["get", "value"], 0, "#f7fbff", 100, "#08306b"]` — the standard MapLibre choropleth ramp — was broken in **WMS `GetMap`** and **Esri `MapServer/export`** alike, since both adapt to the shared `RasterMapRenderingPipeline` / `StyleTranslator` / `SkiaMapRenderer` path.

`InterpolateValues` now types its stops: numeric stops keep the existing lerp expression verbatim, color stops parse to `SKColor` and interpolate per channel including alpha, and anything we cannot evaluate throws instead of defaulting to `0f`.

### What MapLibre actually does, and how that was confirmed

The risk here was replacing "black" with "subtly wrong". Rather than reason from expectations, I installed MapLibre GL JS's own reference implementation (`@maplibre/maplibre-gl-style-spec`) and ran the real expressions through `createExpression(expr, 'fill-color', latest.paint_fill['fill-color'])`, then used its output as the expected values in the tests. They are **not** derived from this implementation.

MapLibre's default `interpolate` (`rgb` space) is a **straight, non-premultiplied per-channel lerp in gamma-encoded sRGB**, including alpha — `Color.interpolate` reads the un-premultiplied `rgb` getter and applies `interpolateNumber` (`from + t * (to - from)`) per channel. Three findings that would each have made a plausible implementation wrong:

| Check | MapLibre output | Rules out |
|---|---|---|
| `#000000` → `#ffffff` @ t=0.5 | `(128,128,128)` | **linear-light** blending (would give ~188) |
| `rgba(255,0,0,1)` → `rgba(0,0,255,0)` @ t=0.5 | `rgba(128,0,128,0.5)` | **premultiplied** lerp (would give `rgba(255,0,0,0.5)`) |
| `#f7fbff` → `#08306b` @ t=0.5, green | `149` (not `150`) | **byte-space** math |

That last one is subtle: MapLibre holds channels as 0..1 float64 and only quantizes at raster time, so `149.5` in byte space is `149.4999…` normalized and rounds down. Channel math therefore runs in normalized 0..1 double precision to match. Perceptual spaces are deliberately not used — MapLibre reserves those for `interpolate-hcl` / `interpolate-lab`, which are separate operators.

Zero-alpha stops also keep their RGB channels (MapLibre preserves the original channels for a zero-alpha color); straight interpolation over `SKColor` reproduces that, so `rgba(255,0,0,0)` → `rgba(0,0,255,1)` @ t=0.5 is purple, not blue.

### Numeric interpolation is unchanged

The numeric branch is the original expression, character for character, and `ConvertToFloat` was refactored into a `TryConvertToFloat` core with identical switch arms in identical order — so `ConvertToFloat(x, 0f)` is by construction `TryConvertToFloat(x, out r) ? r : 0f`. Proven empirically: reverting **only** the `InterpolateValues` body to the pre-fix implementation fails all 20 color/throw tests while **all 7 numeric-interpolate cases still pass**, and those numeric expectations are themselves MapLibre's output (`0, 2.5, 5, 10`).

### Unparseable / mixed stops now fail loudly

Previously they silently became `0f` → black. They now throw `StyleExpressionEvaluationException` naming both offending stop types. This matches MapLibre, which rejects the same expressions at style-validation time (`"Expected color but found number instead."`, `"Could not parse color from value 'not-a-color'"`); because this evaluator parses styles lazily at render time, the equivalent failure has to surface at evaluation. A style we cannot evaluate must not render as a confident black polygon. The message describes stop types only — no server internals.

`ParseColor` keeps its existing transparent/black fallbacks but delegates to a new `TryParseColor` core, so interpolation can distinguish "not a color" from "black" **without a second color parser**.

## Changes Made
- `ExpressionEvaluator.InterpolateValues` — dispatch on stop type: numeric lerp (verbatim), color lerp, or throw. Replaces the unconditional `ConvertToFloat(..., 0f)` coercion that returned a number for color stops.
- `ExpressionEvaluator.InterpolateColors` / `InterpolateChannel` — straight non-premultiplied per-channel sRGB lerp including alpha, in normalized 0..1 double precision to match MapLibre's quantization.
- `ExpressionEvaluator.ParseColor` — refactored onto a new `TryParseColor` core that reports failure instead of substituting black; existing fallbacks preserved and covered by a test. No second color parser.
- `ExpressionEvaluator.ConvertToFloat` — refactored onto `TryConvertToFloat` with identical arms; behavior bit-identical for all existing callers.
- New `StyleExpressionEvaluationException` for lazily-evaluated styles MapLibre would reject at validation time.
- 37 new tests in `ExpressionEvaluatorTests`, with MapLibre-derived expected values.

## Testing
- [x] Architecture tests pass

Covers two-stop ramp, multi-stop ramp, below-first/above-last stop, `rgba()` with alpha, zero-alpha endpoints, exact-stop boundaries (first/interior/last), named colors, gamma-vs-linear-light, numeric unchanged, numeric-string stops, and the mixed/unparseable/null throw paths.

- `ExpressionEvaluatorTests` — 69/69 pass (Release).
- Full `Features.Infrastructure.Rendering` suite — 168/168 pass (Release), no regressions in `StyleTranslator` / `SkiaMapRenderer` / `StyledVectorMapRender`.
- `Honua.Architecture.Tests` — 132/132 pass (Release).
- `dotnet build --configuration Release -p:TreatWarningsAsErrors=true` clean; `dotnet format --verify-no-changes` clean.

**No visual baselines changed** — the repo has no golden-image/visual-regression baselines for map rendering (`RenderGoldenTests` covers Markdown/HTML reporting templates; `tests/baselines/client-compat` are protocol-conformance `.cert.json`). This change does move pixels: areas that rendered black now render their correct ramp color. That is the fix.

## Gate Impact
- [ ] None — no gate impact

Internal rendering logic only — no new routes, so no `EndpointRegistry` / `public-interface-proof.json` / catalog changes.

## Docs or Contract Impact
- [ ] OpenAPI spec changed
- [ ] Documentation updated

## Release/Deploy Impact

## Breaking Changes
None for valid styles. Two intentional behavior changes:
- Styles using color `interpolate` now render correct colors instead of black (the fix — it moves pixels).
- An `interpolate` with mixed-type, unparseable, or null stop outputs now throws rather than silently rendering black. These are styles MapLibre itself rejects at validation time; the previous silent `0f` default is the root cause this PR removes.

## Coordination

- **#2866 (WMS `GetLegendGraphic`) — not touched by this PR.** It reports `interpolate`-over-colors as unrepresentable via a `Warning` header *because `GetMap` could not render it either*. That was the correct call at the time. Now that colors interpolate, that `Warning` may be stale — the legend could represent these as discrete sampled entries. Worth revisiting **after both land**; deliberately left to #2866's owner rather than edited here. Noted on #2867.
- **#2868 (WMS `GetMap` never passes zoom)** is a sibling defect from the same investigation and is explicitly out of scope here.

---

## Pre-PR Checklist

> **Note on `pre-pr-check.sh`:** the script currently fails on Windows before it runs anything, for reasons unrelated to this change — it generates the temporary `.slnf` with forward-slash project paths while `Honua.sln` stores backslashes, so `dotnet restore` aborts with `MSB5028: ... includes project ... that is not in the solution file`. Verified pre-existing: it fails identically on a clean `origin/trunk` checkout with only a whitespace touch (it named a different project, `Honua.Architecture.Tests`). The equivalent checks were run individually instead and all passed — Release build with `TreatWarningsAsErrors=true`, `dotnet format --verify-no-changes`, the affected rendering test suites, and the architecture tests (all listed under Testing above). Filing separately rather than fixing it in this PR.

### DIFF
--- a/src/Honua.Hosting/Features/Rendering/ExpressionEvaluator.cs
+++ b/src/Honua.Hosting/Features/Rendering/ExpressionEvaluator.cs
@@ -2,6 +2,7 @@
 // Licensed under the Elastic License 2.0. See LICENSE in the project root.
 
 using System.Collections.Immutable;
+using System.Diagnostics.CodeAnalysis;
 using System.Globalization;
 using SkiaSharp;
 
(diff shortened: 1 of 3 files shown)


mikemcdougall:

PR #2872, opened 2026-07-15
fix(wms): derive zoom for GetMap so minzoom/maxzoom gates apply (#2868)

mikemcdougall:
# Pull Request

## Issue Link
Closes #2868

## Summary

WMS `GetMap` never supplied a zoom, so `StyleTranslator.ShouldRenderLayer`'s `zoom = null` default silently skipped every layer's `minzoom`/`maxzoom` gate — every layer drew at every scale.

The audit the ticket asked for found **WMS was not the only broken caller — every server-side render path was**. The shared draw path `RasterMapRenderingPipeline.RenderLayerToCanvas` (reached by WMS `GetMap`, Esri `MapServer/export`, the raster tile pipeline, OGC API Maps, static maps, and the bound-style vector compositor) never called `ShouldRenderLayer` at all, so it could not gate even if a zoom had been available. The one path that did call it — `SkiaMapRenderer.RenderWithStyles` — was only ever reached by callers relying on the `null` default. The gate was inert product-wide.

This derives a MapLibre zoom from the rendered envelope and pixel size, threads it through all of those paths, and closes the `null` trap so the next caller cannot silently repeat this.

**How zoom is derived.** MapLibre GL JS defines the mercator world as `Transform.tileSize * 2^zoom` pixels, with `tileSize` a constant 512 (`src/geo/transform_helper.ts`), so `zoom = log2(pixels / (512 * mercatorSpan))`, taking `min` over the two axes exactly as MapLibre's own `cameraForBoxAndBearing` does (`Math.min(scaleX, scaleY)`, `src/geo/projection/camera_helper.ts`). Verified against MapLibre's published `mercatorXfromLng`/`mercatorYfromLat` (`src/geo/mercator_coordinate.ts`): the bounds MapLibre reports for five known camera zoom/center/viewport combinations round-trip back to the original zoom to ~1e-14 (`RenderZoomTests`).

The derivation is deliberately **scale-based, not tile-level-based**: zoom is the camera zoom at which an image is displayed 1:1. This is the only self-consistent rule — the same envelope at the same pixel size now gates identically no matter which protocol asked for it. Two consequences worth naming, both correct rather than incidental:

- A 256px tile of matrix level *z* derives to camera zoom *z − 1*. That is right: a MapLibre client with a 256px raster source sits one camera zoom below the matrix level it fetches, so a style saying "labels above z14" starts drawing them exactly when the camera reaches 14.
- Static maps derive rather than reuse `parameters.Zoom`, because `CenterZoomToExtent` defines that zoom against 256px tiles while MapLibre's camera zoom is against 512px — they differ by one. Deriving also covers the bbox route, which carries no zoom at all.

**Non-WebMercator CRSs.** Gating applies to every CRS whose extent can be expressed in Web Mercator, and it is exact rather than approximate. EPSG:3857 and its aliases derive directly; geographic and other CRSs reproject through the shared transform services the render paths already use. Reprojecting to mercator *before* deriving is what makes this MapLibre-correct for projected CRSs — it naturally accounts for the `1/cos(latitude)` mercator distortion that makes a UTM ground span a different mercator span, which a ground-resolution shortcut would have got wrong by a full zoom level at 60° latitude. Only a CRS with no available transform at all is left ungated, and that path is explicit and traceable rather than silent.

**Closing the trap.** `zoom` is no longer `double? = null`. It is a required `RenderZoom` — a reference type with no public constructor and no meaningful default, so "no zoom" is unreachable by omission and every render path must state which case it is in: `RenderZoom.At(level)` or `RenderZoom.NotDerivable(reason)`. The reason is recorded on the current activity (`honua.render.zoom` / `honua.render.zoom_ungated_reason`), so an ungated render is diagnosable instead of invisible. This is not theoretical — making the parameter required caused the compiler to catch a call site that had been missed while writing this change.

## Changes Made
- Add `RenderZoom` (`src/Honua.Hosting/Features/Rendering/RenderZoom.cs`): a MapLibre zoom derived from an envelope + pixel size, or an explicit `NotDerivable` carrying a reason.
- Add `RasterMapRenderingPipeline.DeriveRenderZoomAsync`, which resolves CRSs with no in-process mercator mapping through the shared coordinate transform services and records the outcome on the current activity.
- Make `StyleTranslator.ShouldRenderLayer`'s `zoom` a required `RenderZoom` instead of `double? = null`, and require it on `SkiaMapRenderer.RenderMap` / `RenderFeaturesOnCanvas` / `RenderWithStyles`.
- Gate `RasterMapRenderingPipeline.RenderLayerToCanvas` on `ShouldRenderLayer`, which it never called — this is what makes gating actually reach WMS, MapServer export, tiles, OGC API Maps, and static maps.
- Derive and pass a zoom from WMS `GetMap` (`BBOX` + `WIDTH`/`HEIGHT` + `CRS`), Esri `MapServer/export`, the raster tile core, OGC API Maps, the bound-style vector compositor, static maps, and print.
- Add `StyleTranslator.IsAnyLayerInZoomRange` and use it in WMS `GetMap` and the tile core to skip *querying* a layer whose style is entirely out of range, not just skip drawing it (the performance tail called out in the ticket).
- Tests: `RenderZoomTests` (MapLibre reference viewports, tile-size semantics, aspect handling, unsupported CRS, degenerate input), `OgcClassicWmsZoomGatingTests` (min/max zoom gating end-to-end over WMS `GetMap`), plus `StyleTranslator`/`SkiaMapRenderer` boundary coverage.
- Regenerate `docs/gis/data/feature-catalog.json` (in Release) — the only change is the three new `[Endpoint]`-attributed proving tests joining the existing WMS route entry. No new routes, so `EndpointRegistry` and `public-interface-proof.json` are untouched.

## Testing
- [x] Architecture tests pass

Boundary semantics are pinned to MapLibre's asymmetric rule — `zoom < minzoom` hides, `zoom >= maxzoom` hides — at 4.99 / 5.0 / 9.99 / 10.0 for a `minzoom:5,maxzoom:10` layer, so the existing `<` / `>=` asymmetry cannot regress.

The new WMS integration tests were verified to actually catch the bug rather than pass vacuously. With the zoom derivation reverted to `NotDerivable` (reproducing trunk's behavior), `Wms_GetMap_MinZoomScopedStyle_OmitsLayerBelowThreshold` and `Wms_GetMap_MaxZoomScopedStyle_OmitsLayerAtOrAboveThreshold` both fail with the layer drawing where it should not; restoring the derivation makes them pass. The unscoped baseline test passes either way by design, so a gating assertion cannot pass merely because the style silently failed to bind.

## Gate Impact

## Docs or Contract Impact

No new routes, so no `EndpointRegistry` / `public-interface-proof.json` / catalog changes.

## Release/Deploy Impact

**This changes rendered output**, which is the point: layers with `minzoom`/`maxzoom` that previously drew at every scale will now disappear outside their declared range. No server-rendered golden-image baselines exist in the repo, so none needed regenerating — the only committed rendered snapshot (`tests/js-browser/esri-leaflet/rendering.spec.ts-snapshots/`) is a client-side Leaflet render of vector features from `/rest/services/.../FeatureServer` and is not produced by this pipeline.

One deliberate, disclosed side effect: routing `RenderLayerToCanvas` through the shared `ShouldRenderLayer` gate also makes it honor `layout.visibility: "none"`, which it previously ignored. That is the same latent defect in the same helper, `SkiaMapRenderer.RenderWithStyles` already honored it, and adding a second zoom-only entry point purely to preserve "draw the layer the cartographer hid" would have been worse than reusing the canonical gate.

## Coordination

Two notes for whoever lands #2866 (WMS `GetLegendGraphic`), which is open and touches `SkiaMapRenderer.cs`:

1. **It will need a one-line update after this lands.** #2866 calls `StyleTranslator.ShouldRenderLayer(styleLayer, zoom)` with a `double? zoom` that is `null` when `SCALE` is absent. That signature is now `RenderZoom`, so its no-SCALE path becomes `RenderZoom.NotDerivable("WMS SCALE parameter was not supplied")` and its SCALE path `RenderZoom.At(...)`. This is the trap-closing working as intended: the null path has to be named rather than defaulted.

2. **Its scale→zoom conversion looks off by one against MapLibre**, and is worth a look before the two ship together. #2866 uses `LegendScaleDenominatorZoom0 = 559082264.0287178` — the GoogleMapsCompatible/WMTS zoom-0 scale denominator, which is derived from **256px** tiles. MapLibre's camera zoom 0 is a **512px** world, i.e. scale denominator `279541132.0143588`. So `log2(559082264.0287178 / SCALE)` returns exactly **MapLibre camera zoom + 1**. If both land as-is, `GetLegendGraphic` and `GetMap` gate the same style at the same physical scale one zoom apart — the legend would list a layer the map does not draw. I have not touched #2866; flagging only.

## Breaking Changes
None — internal APIs only.

---

## Pre-PR Checklist
- [x] Ran `scripts/ci/pre-pr-check.sh` and all checks passed<sup>*</sup>

<sup>*</sup> `scripts/ci/pre-pr-check.sh` is currently broken on a Windows host and aborts at its restore step with `MSB5028: Solution filter file ... includes project ... that is not in the solution file`. This is **pre-existing and unrelated to this PR** — it reproduces identically on a clean `origin/trunk` checkout with no changes applied. The equivalent checks were run directly instead, all green: full-solution Release build with `TreatWarningsAsErrors=true`, `dotnet format Honua.sln --verify-no-changes`, `Honua.Architecture.Tests`, the full WMS suite (65/65), and the affected rendering unit tests.

### DIFF
--- a/docs/gis/data/feature-catalog.json
+++ b/docs/gis/data/feature-catalog.json
@@ -9782,6 +9782,9 @@
         "Honua.Server.Tests.Features.Protocols.Ogc.Classic.Wms.OgcClassicWmsTests.Wms_GetMap_WithoutFilter_NoRegression",
         "Honua.Server.Tests.Features.Protocols.Ogc.Classic.Wms.OgcClassicWmsTests.Wms_InvalidServiceId_ReturnsBadRequest",
         "Honua.Server.Tests.Features.Protocols.Ogc.Classic.Wms.OgcClassicWmsTests.Wms_InvalidService_ReturnsBadRequest",
+        "Honua.Server.Tests.Features.Protocols.Ogc.Classic.Wms.OgcClassicWmsZoomGatingTests.Wms_GetMap_MaxZoomScopedStyle_OmitsLayerAtOrAboveThreshold",
+        "Honua.Server.Tests.Features.Protocols.Ogc.Classic.Wms.OgcClassicWmsZoomGatingTests.Wms_GetMap_MinZoomScopedStyle_OmitsLayerBelowThreshold",
+        "Honua.Server.Tests.Features.Protocols.Ogc.Classic.Wms.OgcClassicWmsZoomGatingTests.Wms_GetMap_UnscopedStyle_RendersAtEveryScale",
         "Honua.Server.Tests.Features.Security.WmsServiceRbacTests.GetCapabilities_WithAnonymousClient_ReturnsWmsAccessDeniedException"
       ],
       "proof_ledger_surface": "map-server",
--- a/src/Honua.Hosting/Features/Rendering/RasterMapRenderingPipeline.cs
+++ b/src/Honua.Hosting/Features/Rendering/RasterMapRenderingPipeline.cs
@@ -287,7 +287,8 @@ internal static class RasterMapRenderingPipeline
                 TileSize,
                 true,
                 null,
-                MetadataV2GeometryType.None);
+                MetadataV2GeometryType.None,
+                RenderZoom.NotDerivable("the tile has no render layers"));
 
             return RasterTileRenderResult.Success(emptyImage, 0);
         }
@@ -311,6 +312,13 @@ internal static class RasterMapRenderingPipeline
         var canvas = surface.Canvas;
         canvas.Clear(SKColors.Transparent);
         var transform = SkiaMapRenderer.BuildTransform(renderExtent, TileSize, TileSize);
+
+        // The tile matrix level is not the MapLibre zoom: these tiles are TileSize (256) pixels, so
+        // a client consuming them as a 256px source sits one camera zoom below the matrix level.
+        // Deriving from the tile envelope keeps the gate aligned with that camera zoom and with the
+        // other protocols rendering the same envelope.
+        var renderZoom = RenderZoom.FromExtent(renderExtent, TileSize, TileSize, tileSrid);
+        RecordRenderZoom(renderZoom);
 
         for (var layerIndex = 0; layerIndex < renderLayers.Count; layerIndex++)
         {
@@ -343,6 +351,11 @@ internal static class RasterMapRenderingPipeline
                 styleCatalog,
                 layer.LayerId,
                 cancellationToken).ConfigureAwait(false);
+            if (!StyleTranslator.IsAnyLayerInZoomRange(stylePlan.StyleLayers, renderZoom))
+            {
+                continue;
+            }
+
             var featureQuery = CreateRasterFeatureQuery(
                 stylePlan,
                 spatialFilter,
@@ -379,7 +392,7 @@ internal static class RasterMapRenderingPipeline
             }
 
             totalFeatureCount += features.Length;
-            RenderLayerToCanvas(canvas, features, stylePlan.StyleLayers, transform, layer.GeometryType);
+            RenderLayerToCanvas(canvas, features, stylePlan.StyleLayers, transform, layer.GeometryType, renderZoom);
         }
 
         var imageBytes = SkiaMapRenderer.EncodeSurface(surface, "png");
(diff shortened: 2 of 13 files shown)


mikemcdougall:

PR #2885, opened 2026-07-16
fix(ci): make jq CR-safe on Windows across scripts/ci (#2876)

mikemcdougall:
# Pull Request

## Issue Link
Closes #2876

## Summary
The standard `jq` for Windows is a native PE binary that writes stdout in **text mode**, so it translates every LF it emits into CRLF, appending a `\r` to each output line. Shell code under `scripts/ci/` captures `jq -r` output into variables and then compares it, tests it for emptiness, uses it as a path, or iterates it — so the trailing `\r` silently corrupts the value. This blocked the pre-PR gate on Windows for `scripts/ci/`-only diffs (behind #2871/#2880) and silently mis-scoped the local build set for others.

Two confirmed symptoms:
- `validate-server-test-binary-artifacts.sh` reported a registered project as **missing** because the captured path carried a trailing `\r` (`[[ -f "$path" ]]` failed).
- `pre-pr-check.sh`'s shard **monolith-fallback** silently died: a shard with an empty `csproj` yielded `"\r"` (non-empty), so `${csproj:-${MONOLITH}}` never fell back; `run_all` (`true\r`/`false\r`) compared correctly only by luck.

Fixed as a **class** with one mechanism: a sourced `jq` wrapper that strips the text-mode CR at line ends. It is the exact inverse of the CRLF translation, is a **no-op on Linux** (CI is unaffected), and preserves any legitimate CR embedded inside a JSON string value.

## Changes Made
- Add `scripts/ci/lib/jq-cr-safe.sh` — a sourced shell fragment defining a `jq()` wrapper: `command jq "$@" | sed 's/\r$//'; return "${PIPESTATUS[0]}"`. It strips ONLY a CR immediately before a line end (`\r$`), so embedded CRs survive and LF-only output passes through unchanged. `PIPESTATUS[0]` preserves jq's own exit status, so `jq -e` truthiness/parse-error codes keep working regardless of `pipefail`. Not exported, so children get the real binary and source it themselves.
- Source the wrapper in the six CR-sensitive scripts on `pre-pr-check.sh`'s execution paths: `pre-pr-check.sh`, `honua-server-targeted-tests.sh`, `validate-ci-router.sh`, `validate-server-test-binary-artifacts.sh`, `validate-server-test-transfer-benchmark.sh`, `validate-server-test-shard-cache.sh`. In scripts with a `command -v jq` / `require_command jq` presence guard, the wrapper is sourced **after** the guard so it still probes the real binary.
- Remove the now-redundant per-call-site `| tr -d '\r'` (2x) and `${shard_name%$'\r'}` patches in `pre-pr-check.sh` so there is a single mechanism instead of scattered partial guards (builds on top of #2880, does not revert it).
- Left `run-server-test-shard.sh` unwrapped deliberately: its single `jq` writes a timing JSON file that is only ever re-parsed as JSON (never shell-string-tested), and its `command -v jq` branch selects a `printf` fallback that a shadowing function would break.

## Testing
- [ ] Unit tests added/updated
- [ ] Architecture tests pass

Verified on Windows (Git Bash, `jq-1.7.1`, PE32+ binary) with the reproduced CR:
- `validate-server-test-binary-artifacts.sh`: **fails before** (`::error::Registered project does not exist: tests/dotnet/Honua.Ai.Tests/Honua.Ai.Tests.csproj`), **passes after** (exit 0).
- `validate-server-test-shard-cache.sh`, `validate-server-test-transfer-benchmark.sh`, `validate-ci-router.sh` (full, 711 test-class ownership assertions): pass.
- `honua-server-targeted-tests.sh`: emits CR-free JSON (`cat -A` shows `}$`, not `}^M$`).
- `pre-pr-check.sh` on this `scripts/ci/`-only diff: **runs to completion** via the CI-shell-only path (steps 1-5), exit 0.
- Wrapper unit proofs: CR stripped (`x^M$` -> `x$`); `jq -e '.missing'` still returns exit 1; embedded `\r` in a value preserved (`a\rb` -> bytes `61 0d 62 0a`, only the trailing text-mode CR removed); LF-only input unchanged (no-op on Linux); empty `csproj` resolves to the monolith fallback; `run_all` compares as clean `true`.

## Gate Impact

## Docs or Contract Impact

## Release/Deploy Impact

## Breaking Changes
None

---

## Pre-PR Checklist
- [ ] Tests added for new functionality

## Notes
- **Scope boundary:** the fix covers the seven scripts the ticket enumerated (all on `pre-pr-check.sh`'s execution paths). `scripts/ci/merge-train/fixtures/validate-merge-train.sh` and the merge-train `lib.sh` it sources share the same CR-sensitive `jq` pattern and are reachable from `pre-pr-check.sh` step 5, but only when the diff touches `scripts/ci/merge-train/`. They run on Linux CI and are out of the ticket's stated scope; recommend a follow-up to source the same wrapper there.
- **Environment note (not part of this fix):** on the test machine, `python3` resolves to the Microsoft Store App Execution Alias stub while `python`/`py` work; the sub-validators (`validate-ci-router.sh`, `check-server-test-shard-coverage.py`, etc.) invoke `python3` directly, so completing `pre-pr-check.sh` required shimming `python3`->`python`. #2880 fixed this class for `pre-pr-check.sh` itself but not these children — a candidate follow-up.


mikemcdougall:

mikemcdougall:

PR #2917, opened 2026-07-17
ci(merge-train): harden dispatch, artifacts, and attribution

mikemcdougall:
﻿## Summary
- Add bounded dispatch/run polling and explicit failure in smart-CI when batch CI never starts (PAT issue) or hangs.
- Regenerate merge-train derived artifacts inside live runs and commit drift before CI dispatch.
- Re-run smart-CI through a single wrapper so regeneration runs consistently.
- Add bounded suspect bisection for real CI failures to isolate culprits before escalate.

Closes #2916, #2915, #2914


mikemcdougall:
This PR is now effectively redundant: ci: harden merge-train dispatch, derived artifacts, and attribution probing (ddc82ff2e) is already on trunk, and this branch has no net diff versus trunk. Marking as closed to reduce noise.

PR #2918, opened 2026-07-17
Bundle MCP auth scheme precedence and session binding

mikemcdougall:
## Summary\n\n- Use authentication-scheme-aware principal keys for MCP session binding\n- Ensure bearer token validation always runs on MCP endpoint when Bearer header is present (even if API key is already in context)\n- Add MCP integration and unit coverage for cross-scheme session behavior\n- Document bearer/API-key precedence and scheme binding\n\nCloses #2851\nCloses #2852


mikemcdougall:
Addressed the remaining unaddressed review feedback from github-code-quality with commit 499e1d0903c47b5bf73ac50d923654b8ab2bd6ee:
- simplified IsScopeGoverned with Any
- simplified CollectRecognizedScopes with Where/SelectMany
- simplified PermitsOperation with Any

Committed on eat/2851-2852-bundle and pushed.

PR #2942, opened 2026-07-20
fix(stac): publish admin layers to catalog

mikemcdougall:
## Issue Link
Closes #2931

## Summary
Standard admin layer publishing now creates a Metadata v2 STAC collection publication alongside the existing FeatureServer publication. STAC search also resolves source-backed readers through the canonical provider router, so newly published layers are visible and queryable through the STAC surface.

## Changes Made
- Create an idempotent `StacCollection` publication in the same Metadata v2 graph save as the FeatureServer publication.
- Preserve protocol-specific publications for the same service/layer instead of replacing one another during upsert.
- Route STAC search through `FeatureProviderQueryRouter` for storage-bound publications, retaining the shared reader fallback for seeded collections.
- Add an integration regression covering admin publish, collection listing/detail, and source-backed item search.

## Testing
- [ ] Unit tests added/updated
- [ ] Architecture tests pass

## Gate Impact
- [ ] None - no gate impact

## Docs or Contract Impact
- [ ] OpenAPI spec changed
- [ ] Documentation updated
- [x] None - no docs or contract impact

## Release/Deploy Impact
- [x] None - standard merge-and-release flow

## Breaking Changes
None

---

## Pre-PR Checklist
- [ ] Ran `scripts/ci/pre-pr-check.sh` and all checks passed

### Validation performed
- `PublishLayer_CreatesDiscoverableStacCollection`: passed.
- `StacSearchTests.SearchGet_WithCollections_FiltersResults`: passed.
- Targeted whitespace formatting for all three touched files: passed.


mikemcdougall (commented):
Actionable finding:

**P1: Route all STAC item reads for source-backed publications, not only search.**

The issue explicitly notes that item-level access works only for collections seeded another way. This patch creates a source-backed `StacCollection` and fixes `/stac/search`, but `ItemEndpoints.HandleGetItems` and `HandleGetItem` still execute through the injected default `IFeatureReader`. The collection document produced here advertises those routes via its `items` links, so a layer published from a secure/external connection can be discoverable and searchable while its canonical item routes read the central/default store, returning empty data, the wrong same-numbered layer, or a 500.

Please resolve the reader from the publication's `StorageBindingId` through `FeatureProviderQueryRouter` in both item handlers (and use the same graph snapshot/visibility result where possible), then extend the integration regression to cover `GET /stac/collections/{id}/items` and `GET /stac/collections/{id}/items/{itemId}`. Until then, `Closes #2931` overstates completion of the reported item-level gap.

src/Honua.Protocols.Stac/SearchEndpoints.cs
```diff
@@ -394,13 +401,28 @@ private static async Task<IResult> ExecuteSearchAcrossPublicationsAsync(
                 var query = layerQueryResult.Query;
                 var projection = layerQueryResult.Projection;
                 var layerId = target.LayerIndex;
+                var targetReader = featureReader;
+                if (snapshot is not null &&
+                    providerQueryRouter is not null &&
```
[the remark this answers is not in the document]
mikemcdougall: Addressed in 0b122c74c. Search now carries the router and snapshot as one nullable routing tuple, removing the redundant always-true router guard while preserving the snapshot and storage-binding gates. `dotnet build src/Honua.Protocols.Stac/Honua.Protocols.Stac.csproj -c Release --no-restore` passes with 0 warnings/errors.

mikemcdougall:
Addressed the P1 source-backed item-routing review in 0b122c74c. Both `GET /stac/collections/{collectionId}/items` and item detail now resolve `FeatureProviderQueryRouter` from the validated V2 publication/resource/service and fall back to the injected reader for seeded publications without storage bindings. The admin-publish regression poisons the canonical default-store row, then verifies search, item list, and item detail still return the source-table feature. STAC protocol project build passes with 0 warnings/errors; the source-backed and seeded test invocations reached test execution but timed out locally due the contended test environment.

src/Honua.Protocols.Stac/SearchEndpoints.cs
```diff
@@ -394,13 +403,27 @@ private static async Task<IResult> ExecuteSearchAcrossPublicationsAsync(
...
+                var targetReader = featureReader;
+                if (providerRouting is { } routing &&
+                    !string.IsNullOrEmpty(target.Publication.StorageBindingId))
+                {
+                    targetReader = await routing.Router.ResolveReaderAsync(
+                            routing.Snapshot,
+                            target.Service,
+                            target.Resource,
+                            target.Publication,
+                            layerId,
```
mikemcdougall: [P1] Carry the binding's storage handle through routing and execution. `target.LayerIndex` is the protocol-facing/service-local index, while `FeatureProviderQueryRouter.ResolveReaderAsync` explicitly expects the provider `storageLayerId`; `MetadataV2GraphSnapshotExtensions.ResolveStorageLayerId` documents that these values are distinct. Several bindable providers also reject a query whose `layerId` differs from `binding.StorageLayerId`. This currently resolves and then calls `CountAsync`/`QueryAsync` with `target.LayerIndex`, so a valid publication with different public and storage IDs fails or addresses the wrong provider layer. The admin regression uses equal IDs and cannot catch it. Please resolve `routing.Snapshot.ResolveStorageLayerId(target.Publication)` and carry both the reader and resolved storage ID through every provider call; the item list/detail path has the same issue for `QueryAsync`/`GetAsync`.

src/Honua.Protocols.Stac/ItemEndpoints.cs
```diff
@@ -304,6 +323,34 @@ private static string BuildItemsFilterQuery(int limit, int offset, string? bbox,
...
+        HttpContext context,
+        IFeatureReader featureReader,
+        MetadataV2Service? service,
+        MetadataV2Resource resource,
+        MetadataV2Publication publication,
+        int layerId,
+        CancellationToken cancellationToken)
+    {
+        var providerQueryRouter = context.RequestServices.GetService<FeatureProviderQueryRouter>();
+        if (providerQueryRouter is null || service is null || string.IsNullOrEmpty(publication.StorageBindingId))
```
mikemcdougall: [P1] Bound publications should fail closed when provider routing is unavailable. Returning the shared `IFeatureReader` here is only safe for seeded/unbound collections; for a publication with `StorageBindingId`, a missing router silently reads the canonical store using the same public layer index and can return unrelated data across a provider or tenant boundary. The search path has the same broad optional-router fallback. Please restrict fallback to publications without a storage binding and require provider routing (or return a controlled server error) for bound publications. Add a regression where the bound source and default store contain conflicting rows and routing is unavailable/misconfigured.

mikemcdougall:
Review summary at `0b122c74c`:

- Two P1 provider-routing/isolation findings are left inline.
- Authorization filtering itself remains on the shared STAC lookup/validation paths; I did not find a new direct visibility bypass in the changed code.
- Regression coverage should include `Publication.LayerIndex != StorageBinding.StorageLayerId`, fail-closed behavior for a bound publication without routing, and a restricted/tenant-hidden publication. The current source-poisoning test proves the happy-path router selection only when both IDs are equal and the full DI graph is present.
- Issue #2931's second acceptance criterion (remove the `honua-samples/samples/stac-search-rest` workaround note) is not represented in this PR. Either land/cross-link that samples change before closing #2931 or avoid `Closes #2931` until it is complete.
- The current PR Gate failure is the reported generated feature-catalog drift; the fast server tests themselves passed 2821/2821 and architecture tests were 173/174 with only `CommittedCatalog_EqualsFreshlyGeneratedOutput` failing.

No approval or merge action taken.

src/Honua.Protocols.Stac/ItemEndpoints.cs
```diff
@@ -248,22 +259,49 @@ private static async Task<IResult> HandleGetItem(
...
+            var isStorageBound = !string.IsNullOrEmpty(publication.StorageBindingId);
+            var hasNumericItemId = long.TryParse(
                 itemId,
-                cancellationToken);
-            if (feature is null && long.TryParse(itemId, NumberStyles.Integer, CultureInfo.InvariantCulture, out var objectId))
+                NumberStyles.Integer,
+                CultureInfo.InvariantCulture,
+                out var objectId);
+            Feature? feature;
+            if (isStorageBound && hasNumericItemId)
```
mikemcdougall: **[P1] Do not equate a numeric STAC item ID with the provider object ID.** `StacMappingService.ResolveItemId` deliberately prefers `stac_id`, then `item_id`, then `id`, and those attributes can be numeric while differing from `Feature.ObjectId`. The item list will therefore emit (for example) a self-link ending in `/items/123` for a feature whose object ID is 900, but this bound path now fetches object ID 123 and can return the wrong feature (or 404). The direct fetch also never verifies that the mapped item ID matches the requested ID. Please preserve canonical-ID precedence for bound providers using a provider-neutral field lookup, with object-ID fallback only after canonical lookup; at minimum never return an object-ID hit whose `ResolveItemId` differs. Add a regression with numeric `id`/`stac_id` distinct from object ID plus a colliding object ID so list/search/detail identity is proven end to end.

mikemcdougall (commented):
The original storage-handle and fail-closed P1s are addressed, but the final item-detail change introduces one actionable regression.

src/Honua.Protocols.Stac/ItemEndpoints.cs
```diff
@@ -248,18 +261,57 @@ private static async Task<IResult> HandleGetItem(
...
+                    cancellationToken);
+                if (feature is null && hasNumericItemId)
+                {
+                    feature = await TryGetFeatureByObjectIdAsStacAsync(
+                        readerResolution.Reader,
+                        readerResolution.StorageLayerId,
+                        resourceSrid,
+                        objectId,
+                        cancellationToken);
+                }
```
mikemcdougall, line 306: **[P1] Verify the effective mapped STAC ID before returning either canonical or OID candidates.** The new loop accepts `GetCanonicalItemMatchRank != null`, which means a lower-precedence field can match even when a higher-precedence field determines the actual response ID. With the regression fixture feature (`stac_id=123`, `id=900`), `/items/900` queries `id=900`, accepts that feature, then `MapFeatureToItem` returns `id: 123`. Separately, if no canonical query matches, the OID fallback still returns object ID N even when that feature has `stac_id`/`item_id`/`id` mapping to a different value. Both cases violate the self-link identity and can return the wrong record. Share or replicate `ResolveItemId` and require its effective value to equal `itemId`; allow OID fallback only when that equality holds, including the no-canonical-attribute case where effective ID is OID. Add negative coverage for `/items/900` in the current fixture and for an OID collision with a different effective canonical ID.

mikemcdougall (commented):
Provider-neutral query translation is in place, but the canonical-ID invariant is still not enforced on the selected record.

src/Honua.Protocols.Stac/SearchEndpoints.cs
mikemcdougall: Addressed at exact head 7fbcf42eb: StacFeatureReaderResolver now returns the resolved binding StorageLayerId, and search/list/detail pass that storage ID to CountAsync/QueryAsync while retaining the public LayerIndex for STAC mapping. The integration regression binds the publication to a deliberately different storage layer ID and verifies source-backed search, list, and detail results.

src/Honua.Protocols.Stac/ItemEndpoints.cs
mikemcdougall: Addressed at exact head 7fbcf42eb: bound STAC publications now fail closed when routing dependencies, the resource graph, the service, the storage binding, or the bound reader cannot be resolved. The integration regression removes the resource bindings and assigns an explicit missing binding, then asserts HTTP 500 for search, list, and detail rather than falling back to the default store.

src/Honua.Protocols.Stac/ItemEndpoints.cs
mikemcdougall: Addressed at exact head 7fbcf42eb: detail lookup no longer treats a numeric route ID as an object-ID shortcut. It queries declared canonical fields in stac_id/item_id/id precedence through the provider-neutral typed filter pipeline. The regression uses id=900, stac_id=123 alongside an unrelated object ID 123 and verifies /items/123 returns the canonical feature.

src/Honua.Protocols.Stac/ItemEndpoints.cs
mikemcdougall: Addressed at exact head 7fbcf42eb: every canonical-field and object-ID candidate is accepted only when StacMappingService.ResolveItemId(candidate) exactly matches the requested route ID. Regression assertions verify /items/900 and /items/999 return 404 when the effective mapped STAC ID does not match. Protocol build passed and independent exact-head review reported clean.

src/Honua.Protocols.Stac/ItemEndpoints.cs
```diff
@@ -355,6 +416,81 @@ private static string BuildItemsFilterQuery(int limit, int offset, string? bbox,
...
+                Limit = 2
+            };
+            var result = await featureReader.QueryAsync(storageLayerId, query, cancellationToken);
+            foreach (var candidate in result.Features)
+            {
+                if (GetCanonicalItemMatchRank(candidate, itemId) is not null)
+                {
+                    return candidate;
+                }
+            }
```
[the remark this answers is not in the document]
mikemcdougall: Reviewed at exact head 7fbcf42eb. This loop intentionally performs ordered, short-circuit candidate validation and returns the first candidate whose effective mapped STAC ID matches. Keeping the predicate adjacent to the early return makes the precedence and identity guard explicit; introducing Where would not change behavior or address a correctness issue. Exact-head CI is fully green.

src/Honua.Protocols.Stac/SearchEndpoints.cs
```diff
@@ -431,7 +441,7 @@ private static async Task<IResult> ExecuteSearchAcrossPublicationsAsync(
                     var remaining = effectiveLimit - allItems.Count;
                     query = query with { Limit = remaining };
 
-                    var result = await featureReader.QueryAsync(layerId, query, cancellationToken);
+                    var result = await targetReader.QueryAsync(storageLayerId, query, cancellationToken);
```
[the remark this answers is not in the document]
mikemcdougall: Addressed at exact head 6ec24adcc. Storage-bound STAC ids filters are now built as metadata-typed FilterExpression trees and translated through IQueryProcessor, with stac_id/item_id/id precedence guarded by higher-priority null checks. Routed queries no longer emit the default-store objectid SQL fallback; unbound collections retain the legacy behavior. The focused publishing integration regression passes and proves ids=123 returns exactly the feature whose effective STAC ID is 123 despite a colliding provider PK 123/stac_id 456, while ids=900 returns zero.

src/Honua.Protocols.Stac/SearchEndpoints.cs
```diff
@@ -1430,6 +1456,91 @@ private static SqlFragment BuildItemIdsSqlFilter(ImmutableArray<string> itemIds)
...
+                FilterExpression matches = literals.Length == 1
+                    ? new BinaryExpression(property, BinaryOperator.Equal, literals[0])
+                    : new BinaryExpression(property, BinaryOperator.In, new ValueList(literals));
+                if (higherPrecedenceFieldsAreNull is not null)
+                {
+                    matches = new BinaryExpression(
+                        higherPrecedenceFieldsAreNull,
+                        BinaryOperator.And,
+                        matches);
+                }
```
mikemcdougall: **[P2] Preserve ObjectId fallback and the same absence semantics as `ResolveItemId`.** For a bound resource whose primary key is `fid`/`gid` (no declared `stac_id`, `item_id`, or `id` field), item mapping falls back to `Feature.ObjectId`, but this builder leaves `combined` null and returns `1 = 0`; `/stac/search?ids=<the ID emitted by item list/detail>` can never match. The fallback needs to target `resource.FindPrimaryIdField()` through the same typed `IQueryProcessor` path, gated by all canonical fields being absent, rather than reintroducing bare `objectid` SQL. Also, `ResolveItemId` treats null, empty, and whitespace canonical values as absent, while these precedence predicates require only `IS NULL`; a row with `stac_id=''` and `item_id='x'` maps to `x` but is excluded by `ids=x`. Build the precedence guard to mirror the effective-ID resolver exactly. Add bound regressions for a noncanonical primary-key name/no canonical fields, multiple IDs, and an empty higher-precedence string.

mikemcdougall (commented):
The provider-neutral rewrite removes the unsafe SQL, but it does not yet preserve all effective STAC ID semantics.

src/Honua.Protocols.Stac/SearchEndpoints.cs
mikemcdougall: Addressed at exact head cf5b1b495. The bound ids AST now mirrors ResolveItemId precedence: string canonical fields are absent when NULL or TRIM(field) is empty, lower candidates are guarded by all higher fields being absent, and resource.FindPrimaryIdField() is appended as the final metadata-typed ObjectId fallback without bare objectid SQL. Multiple ids remain typed IN predicates. TRIM translation is now explicitly supported and tested for MySQL and Databricks as well as the existing PostgreSQL path. The expanded focused integration test passes with whitespace stac_id falling through to item_id, combined ids, and a separately published fid primary-key table with no canonical fields. Validation: STAC focused integration 1/1 passed; MySQL translator suite 41/41 passed; Databricks focused TRIM test 1/1 passed.

src/Honua.Protocols.Stac/SearchEndpoints.cs
```diff
@@ -1430,6 +1456,141 @@ private static SqlFragment BuildItemIdsSqlFilter(ImmutableArray<string> itemIds)
...
+
+                combined = combined is null
+                    ? primaryMatches
+                    : new BinaryExpression(combined, BinaryOperator.Or, primaryMatches);
+            }
+        }
+
+        if (combined is null)
+        {
+            return new SqlFragment("1 = 0", []);
```
mikemcdougall: **[P1] Translate with the selected provider, not the process-wide `IQueryProcessor`.** This method turns the AST into a dialect-specific `SqlFilter` before `StacFeatureReaderResolver`/`FeatureProviderQueryRouter` chooses the bound reader. In the default Postgres host that means a Postgres fragment is handed to any routed secondary provider. SQL Server is a concrete registered, bindable provider and `SqlServerFeatureQueryBuilder.AppendWhereClause` explicitly throws for every `FeatureQuery.SqlFilter` because the shared translator emits Postgres SQL; Oracle documents the same restriction, and the bindable Redshift/Snowflake/ArcGIS paths also need an explicit compatibility decision. The new MySQL and Databricks translator unit tests only prove those translators in isolation; they do not prove a Postgres-default host routes this expression through them. Thus `/stac/search?...&ids=` still 500s for relevant bound providers. Carry the provider-neutral `FilterExpression` to provider selection and translate with that provider's translator (or add an equivalent routed query abstraction), then add an end-to-end/mixed-provider contract test for every provider the router can select. If some registered providers intentionally cannot support this filter, fail with a documented capability response rather than claiming provider-neutral support.

mikemcdougall (commented):
Effective-ID precedence is fixed, but provider selection still occurs after dialect-specific translation.

src/Honua.Protocols.Stac/SearchEndpoints.cs
mikemcdougall: Addressed at exact head bb8b5908f. Storage-bound STAC id lookup no longer uses the process-wide IQueryProcessor or carries a dialect-specific SqlFilter through routing. Search/detail now resolve the bound reader first, issue at most four provider-neutral simple typed equality/IN candidate queries for declared stac_id/item_id/id/primary fields, dedupe by object ID, and enforce StacMappingService.ResolveItemId membership before pagination/counting. Candidate overflow fails closed. Narrow parameterized IN support and focused contracts cover Postgres, SQL Server, Oracle, Snowflake, Redshift, and ArcGIS REST; routed-reader tests reject every non-null SqlFilter and cover lower-precedence collisions. Validation: protocol build 0 warnings/errors, provider contracts 6/6, routed/builder contracts 5/5, real PostGIS publish/router/STAC E2E 1/1.

PR #3015, opened 2026-07-24
fix(geocoding): provider config records bind empty under native AOT (init-only properties)

mikemcdougall:
Root cause of demo.honua.io geocoding failing with `Amazon Location place index name is required` despite fully-set `Geocoding__Providers__AmazonLocation__*` env (#2948 follow-through): the provider configuration records use **init-only properties**, which the configuration binding source generator (`EnableConfigurationBindingGenerator`, required for the AOT Lambda image) cannot set — so `AddOptions<T>().Bind(section)` silently produces an empty configuration on AOT builds. Nominatim only ever worked because its code defaults sufficed with zero binding; any env-based geocoding provider configuration has been a silent no-op on AOT images.

## Change
- `GeocodeProviderConfiguration` (base) + `NominatimProviderConfiguration` / `AmazonLocationProviderConfiguration` / `AzureMapsProviderConfiguration`: `init` → `set`. No `with`-expressions exist against these types; record equality and all call sites unchanged.
- New `GeocodingProviderConfigurationBindingTests`: binds each provider config from in-memory configuration through the exact `AddOptions().Bind()` path `ProviderIntegrationExtensions` uses and asserts the properties land.

## Verification (foreground, exit-code gated)
- `dotnet build` (Honua.Geocoding + Honua.Server.Tests, Release): exit 0
- `dotnet format --verify-no-changes` on changed files: exit 0
- New binding tests + `NominatimGeocodeProviderTests` + `GeocodingOptionsValidatorTests`: 9/9, exit 0
- `GeocodingEndpointTests` full class: first run 60/61 with one failure under heavy shared-box contention (14m37s wall clock); re-run of the same class surfaced no failing test through a Failed/Assert/error filter. Reporting both honestly — PR Gate is the verification of record for the box-environmental class per repo precedent.

Demo-side context: the demo now serves v35 with geocoding env set and the `geo.places` PrivateLink endpoint live (honua-io/honua-demo); geocoding end-to-end completes when the next demo image includes this fix.

Refs #2948 (S4; demo image redeploy remains)


mikemcdougall:
@codex review

